PR #1, opened 2023-10-03
add aspect-cli lint command

jbedard:


plugin.go
```diff
@@ -0,0 +1,186 @@
...
+				}
+
+				return nil
+			},
+		),
+	}, nil
+}
+
+func (plugin *LintPlugin) findLintResultFiles(streams ioutils.Streams, bazelStartupArgs []string) ([]string, error) {
+	// TODO: use bazel query
```
alexeagle, line 106: actually the TODO should be to watch the build event stream as it goes by, I expect linting actions along with their output paths (including for cache hits) will be reported.

alexeagle (approved):

PR #2, opened 2023-10-11
better demo with buf linting protos too

alexeagle:


.bazelrc
```diff
@@ -3,6 +3,8 @@
 # Settings that apply only to CI are in .github/workflows/ci.bazelrc
 
 common --enable_bzlmod
+# PR 997 not merged yet
```
jbedard, line 6: Can we put a link?

alexeagle: it will merge in a couple mins, will remove before merging this

example/.aspect/cli/config.yaml
```diff
@@ -1,5 +1,9 @@
 plugins:
   - name: lint-plugin
     # For this to work, you must first `bazel build dev` in the root of the repo
-    from: ../../bazel-bin/plugin
+    from: ../bazel-bin/plugin
     log_level: warn
+    properties:
+      aspects:
```
jbedard, line 7: This is being parsed into a `LintAspects` property, why not name it the same here?

alexeagle: I was on the fence which name to use. WDYT? `lint_aspects` here, or `Aspects` in the Go code?

plugin.go
```diff
@@ -36,6 +39,17 @@ type LintPlugin struct {
 	aspectplugin.Base
 	// This plugin will store some state from the Build Events for use at the end of the build.
 	command_line.CommandLine
+	// Helper to parse our config section
+	yamlUnmarshalStrict    func(in []byte, out interface{}) (err error)
```
jbedard, line 43: Why inject it like this instead of just using the library?

alexeagle: I dunno, I copied from https://github.com/aspect-build/plugin-augment-error/blob/main/plugin.go#L30
Maybe it's meant to be for testability?

example/.aspect/cli/config.yaml
jbedard: For the quickest solution to avoid thinking about that... I'd say `lint_aspects` describes exactly what it is so I'd prefer that out of those 2 options.

But will the person writing this file always be the one writing the aspects? Do we even need the word `aspect`? I guess we're also exposing the `%` syntax here so idk if it's worth thinking about too much atm.

jbedard: I guess we are just blindly inserting these as `--aspects` so in theory someone could pass any aspect here and it might have nothing to do with linting, although we might not want to "allow" that.

It could be something such as:
```yasml
linters:
    eslint@//:lint.bzl
```
or some magic like that to hide all "aspect" logic from this file.

alexeagle: I don't think we're going to hide it that well. You also have to write a `lint.bzl` file to declare the aspects we use. So I think it's okay `lint_aspects`

plugin.go
jbedard: I'd drop it if you had no specific reason. Can always add it back if actually needed 🤷

example/.aspect/cli/config.yaml
jbedard: I just wonder if it would be different people writing aspects vs deciding which to use? `lint_aspects` is clearest atm though 👍

plugin.go
alexeagle: Okay I simplified it to empty "constructor" call

jbedard (approved):

PR #5, opened 2023-10-12
Flake8

alexeagle:
These three commits should not be squashed: they represent the three steps I'll document for adding your own linter.


lint/README.md
```diff
@@ -0,0 +1,80 @@
...
+Run `bazel run docs:update` to create the Markdown file.
+
+Also add your new linter to the README.
+
+Here's a commit showing what it should look like:
+https://github.com/aspect-build/rules_lint/commit/c3bf01a39c2e68b0b37918620aeafbf8ef0b2d85
+
+## Step 4: Send the PR!
+
+We'd love to make your linter available to everyone.
```
kormide, line 80: nitty nit: end with `!`?

lint/flake8.bzl
```diff
@@ -0,0 +1,93 @@
...
+                pkg = "@pip//flake8:pkg",
+            )
+            ```
+        config: the flake8 config file (`setup.cfg`, `tox.ini`, or `.flake8`)
+    """
+    return aspect(
+        implementation = _flake8_aspect_impl,
+        # Edges we need to walk up the graph from the selected targets.
+        # Needed for linters that need semantic information like transitive type declarations.
+        # attr_aspects = ["deps"],
```
kormide, line 81: Remove commented line?

kormide (approved):

lint/flake8.bzl
alexeagle: It's intentionally commented, this commit is also an example, so someone studying it may need this (if they wanted to add mypy, say ... @dzbarsky ...)

PR #7, opened 2023-10-12
chore: tighten up our bzlmod deps

alexeagle:
Also green up main CI by dropping windows for now
(it fails because we don't have an Aspect CLI windows binary yet)

### DIFF
--- a/example/MODULE.bazel
+++ b/example/MODULE.bazel
@@ -1,30 +1,23 @@
 "Bazel dependencies"
 
-bazel_dep(name = "aspect_rules_lint", version = "0.0.0", dev_dependency = True)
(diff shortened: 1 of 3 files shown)


example/MODULE.bazel
```diff
@@ -1,30 +1,23 @@
 "Bazel dependencies"
 
-bazel_dep(name = "aspect_rules_lint", version = "0.0.0", dev_dependency = True)
```
alexeagle, line 3: note: dev_dependency is technically better here, but I just want to reduce the size of the example to make it less intimidating for casual readers

kormide (approved):

PR #9, opened 2023-10-12
suggestion demo

alexeagle:


example/src/lint.ts
```diff
@@ -0,0 +1,3 @@
+// this is a linting violation
+const lint: string = "warning";
```
alexeagle, line 2: [ESLint] Type string trivially inferred from a string literal, remove type annotation  [@typescript-eslint/no-inferrable-types]

```suggestion
const lint = "warning";
```

PR #13, opened 2023-10-13
feat: print nicer output from plugin

alexeagle:
Demo: [asciicast](https://asciinema.org/a/xQWU1Wc1JINOubeguDDQbBqcq)


jbedard (approved):
Nice 👍

PR #14, opened 2023-10-14
feat: merge bazel-super-formatter repo here

alexeagle:
Follow this approach where the tools must be declared in userland

Potentially replaces https://github.com/aspect-build/bazel-super-formatter/pull/49 and results in us archiving bazel-super-formatter :(

/cc @jsharpe @agluszak


jbedard:
Was this decided on or just a POC? Was there a discussion somewhere?

alexeagle:
The idea came from a PR on the super-formatter to add bzlmod. Then it just magically fell together in a couple hours. Let's discuss 🥹

agluszak (commented):
LGTM source-wise, I can help with migrating remaining tools. One of my interns wanted to add rustfmt as well. 

I'll test how it works in practice on Monday.

What about rules_jvm_external and scalafmt?

And what do you think about my idea of testing bzlmod/workspace correspondence?

alexeagle:
Adding rules_jvm_external here is fine since it goes in the users project. I'm sorry you went to all the trouble to vendor individual jar files in the other project!

alexeagle:
My idea is to support only bzlmod here and only workspace in bazel-super-formatter.. can you say more about your idea for correspondence?

docs/linting.md
```diff
@@ -0,0 +1,72 @@
...
+See the `docs/` folder for "aspect factory functions" that declare your linters.
+
+Finally, reference those linters as `//path/to:lint.bzl%mylinter`
+in the lint runner.
+
+If you use the Aspect CLI, then include a block like the following in `.aspect/cli/config.yaml`:
+
+```yaml
+plugins:
+  - name: lint-plugin
```
gregmagolan, line 24: missing `from:` ?

example/package.json
```diff
@@ -3,6 +3,20 @@
     "eslint": "*",
     "@typescript-eslint/parser": "*",
     "@typescript-eslint/eslint-plugin": "*",
-    "typescript": "5.0"
+    "typescript": "5.0",
+    "prettier": "^2.8.7",
+    "prettier-plugin-sh": "^0.12.8",
+    "prettier-plugin-sql": "^0.14.0"
+  },
+  "//FIXME(alexeagle)": "This section shouldn't be needed since these are plugins",
```
gregmagolan, line 11: not needed anymore since you're public hoisting?

example/src/hello copy.tf
```diff
@@ -0,0 +1,8 @@
+resource "aws_subnet" "private_subnet_us_east_2a" {
```
gregmagolan, line 1: `hello copy.tf` snuck in?

example/tools/BUILD
```diff
@@ -0,0 +1,87 @@
...
+    # https://github.com/bazelbuild/rules_swift/issues/864
+    actual = select({
+        "@bazel_tools//src/conditions:linux": "@swiftformat",
+        "@bazel_tools//src/conditions:darwin": "@swiftformat_mac",
+    }),
+)
+
+multi_formatter_binary(
+    name = "format",
+    formatters = {
```
gregmagolan, line 75: 🧙

gregmagolan (approved):
🦖

example/package.json
alexeagle: I couldn't get it to work without this... as usual

docs/linting.md
alexeagle: yeah... but we haven't published the plugin on a release yet, so there's not a good `from` value to document...

alexeagle: it will eventually just be `from: rules_lint` because we default to github.com/aspect-build

example/package.json
alexeagle: come help me later after merge 😁

example/tools/BUILD
alexeagle: actually a dict was a bad idea, better to have one attribute per language so the list of languages doesn't need to be documented

PR #25, opened 2023-10-19
Better workaround for eslint bug

alexeagle:
Changing to ctx.actions.run_shell caused the RUNFILES lookup logic to fail with bzlmod disabled.
(and we're missing test coverage)

### Type of change

### Test plan

- New test cases added
added a "happy path" eslint_test target.


gregmagolan (approved):
🌮

PR #26, opened 2023-10-19
chore: delete the Aspect CLI plugin

alexeagle:
With that code removed, we're okay with licensing this as Apache 2.0

This assumes we'll add the `lint` command in the aspect-cli repo as a built-in command immediately.


docs/linting.md
```diff
@@ -21,10 +21,8 @@ If you use the Aspect CLI, then include a block like the following in `.aspect/c
 
 ```yaml
 plugins:
```
gregmagolan, line 23: remove this bit about plugins entirely?

gregmagolan: ohhh. you meant this to be
```
lint:
  aspects:
    - //tools:lint.bzl%eslint
```

lint/BUILD.bazel
```diff
@@ -21,7 +21,6 @@ bzl_library(
 
 bzl_library(
     name = "lint_test",
```
gregmagolan, line 23: bzl_library with no srcs?

gregmagolan (approved):
🦖

lint/BUILD.bazel
alexeagle: I ran gazelle and it disappointed :/

PR #33, opened 2023-10-26
Replace Black with Ruff as the Python formatter

mortenmj:
Ruff now acts as a Black-compatible formatter. See https://astral.sh/blog/the-ruff-formatter
If it's anything like the rest of Ruff, it should be super fast, so I suggest we use it :rocket:

I moved the ruff binary from examples/ to examples/tools/, but I'm not really sure where these things should live when they're used as both linters and formatters. I also turned it into an alias, from a native_binary, because I didn't see why it should be a native_binary.

The existing versions of ruff complained that the `format` subcommand was experimental, so I bumped to the latest version as it does not say this.


alexeagle (approved):
This is amazing good timing, I was just about to do this 😉

Thank you!!

PR #36, opened 2023-10-31
chore: reference latest aspect CLI version from .bazeliskrc

plobsing:
`aspect/5.7.2` doesn't seem to include the `lint` command and `bazel lint src:all` fails with

```
Error: unknown command "lint" for "aspect"
```

This error is resolved by using version 5.8.5.


alexeagle (approved):
Thanks for all the fixes! Was distracted with conferences last week.

PR #48, opened 2023-11-09
fix: template the fix target string in failure error message

mattem:
There's a number of assumptions that the fix target is `//tools:format` when this might not be the case.
Adds a template string to the fix target in the failure error message on formatting.

---

### Type of change

Ensure formatting error fix target correctly reflects the name of the formatter binary target.

### Test plan

Run the formatter locally with a different target name.


alexeagle (approved):
Nice, thanks!

PR #49, opened 2023-11-09
feat: auto-mirror ruff releases

alexeagle:
If this pattern works well, I'll roll out to all our rulesets that have release mirroring


lint/mirror_ruff.sh
```diff
@@ -31,24 +31,35 @@ map(
...
   --header "Accept: application/vnd.github.v3+json" \
-  https://api.github.com/repos/${REPOSITORY}/releases?per_page=2  
+  https://api.github.com/repos/${REPOSITORY}/releases?per_page=1
 
 jq "$JQ_FILTER" <$RELEASES >$RAW
 
+# Combine the new versions with the existing ones.
+# New versions should appear first, but existing content should overwrite new
+CURRENT=$(mktemp)
+python3 -c "import json; exec(open('$SCRIPT_DIR/ruff_versions.bzl').read()); print(json.dumps(RUFF_VERSIONS))" > $CURRENT
```
thesayyn, line 44: nit: ruff_versions.bzl is a valid json without the prefix. maybe no python-use to minimize these dependencies?

thesayyn (approved):

lint/mirror_ruff.sh
alexeagle: Yeah, I tried that approach for a while but failed to get the right `awk` command, I'll try again as I agree it's hard to know python3 will be on the GHA runner

alexeagle: UGH the trailing comma - it's not actually valid json

Issue #18, opened 2023-10-16
[FR]: Apply fixes


alexeagle:
### What is the current behavior?

Linters offer to fix problems they find, but we don't offer a convenient workflow under Bazel to accept those.

### Describe the feature

Likely https://github.com/aspect-build/aspect-cli/issues/409 is the nicest developer experience.

For developers who don't want to use Aspect CLI we can follow the model of `example/lint.sh` and just show them how to write a `fixup.sh` that does approx. the same thing.

alexeagle:
Started thinking about this.

Tools expect to overwrite their source files, and we don't want to be in the business of modifying tools. So we can't run fix as an action, where Bazel makes inputs read-only.

We can follow the model of format where each lint action produces a 'Bazel run' command that does the fixing for one target.

But tools like eslint don't currently work this way because the runfiles isn't linked the same.

alexeagle:
As a demo, this command successfully applies a fix:

```
% BAZEL_BINDIR="." bazel run --run_under="cd $PWD &&" -- tools:eslint --no-eslintrc --config .eslintrc.cjs --fix src/file.ts
```

but only after I run `pnpm install` in the source tree so the plugin "@typescript-eslint/eslint-plugin" can be located.

Another downside of this approach is we won't have bazel sandbox so the tool might behave differently under `bazel run` than it did when the aspect ran it as an action.

PR #53, opened 2023-11-12
feat(eslint): print the fix command

alexeagle:
Part of #18 but needs more design thought.

Issues:
- you must `pnpm i` in the source folder, in order for eslint to resolve its plugins like `@typescript-eslint/parser`. We already had to declare a proper `js_library(name="eslintrc")` but this doesn't use it since you just `bazel run` the bare eslint binary.
- to avoid polluting the report with instructions, I print this to stderr in the formatter, so the output just looks like
```
alexeagle@MacBook-Pro-8 example % ./lint.sh src:ts
INFO: Analyzed target //src:ts (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
INFO: From ESLint src/ESLint.ts.aspect_rules_lint.report:
1 error and 0 warnings potentially fixable by running
    BAZEL_BINDIR="." bazel run --run_under="cd $PWD &&" -- @//tools:eslint --fix --no-eslintrc --config .eslintrc.cjs src/file.ts
INFO: Elapsed time: 1.359s, Critical Path: 1.23s
INFO: 2 processes: 1 internal, 1 darwin-sandbox.
INFO: Build completed successfully, 2 total actions
```
- the `bazel fixup` command we want to add in the CLI shouldn't have to screen-scrape output. We should have another way to propagate this, it could be anything that's readable from the BES - which probably means it has to be another output file?

---

### Type of change

### Test plan

Ran `./lint.sh src:ts` in the example folder and then ran the fix command it produced

### DIFF
--- a/lint/eslint.bazel-formatter.js
+++ b/lint/eslint.bazel-formatter.js
@@ -16,6 +16,7 @@ function pluralize(word, count) {
 
 module.exports = function (results, context) {
   let output = "";
+  (fixableErrorCount = 0), (fixableWarningCount = 0);
 
   results.forEach((result) => {
     const messages = result.messages;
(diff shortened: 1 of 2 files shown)


alexeagle:
We want to do something more like #62 instead - the lint action should produce a patch file, then applying that patch file is an orthogonal concern that no longer needs to be able to run the linter.

PR #60, opened 2023-11-29
feat: allow multiple ruff config files

alexeagle:
Allows the common monorepo scenario where one application has to have a different configuration for whatever reason.

As documented at https://docs.astral.sh/ruff/configuration/#config-file-discovery the tool already knows how to select the right file for a given python source file.

### DIFF
--- a/lint/ruff.bzl
+++ b/lint/ruff.bzl
@@ -30,14 +30,13 @@ def ruff_action(ctx, executable, srcs, config, report, use_exit_code = False):
         report: output file to generate
         use_exit_code: whether to fail the build when a lint violation is reported
     """
-    inputs = srcs + [config]
+    inputs = srcs + config
     outputs = [report]
 
     # Wire command-line options, see
     # `ruff help check` to see available options
     args = ctx.actions.args()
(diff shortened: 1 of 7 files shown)


lint/ruff.bzl
```diff
@@ -22,22 +22,34 @@ _MNEMONIC = "ruff"
...
     """
-    inputs = srcs + [config]
+    inputs = srcs + config
     outputs = [report]
 
     # Wire command-line options, see
     # `ruff help check` to see available options
     args = ctx.actions.args()
     args.add("check")
-    args.add(config, format = "--config=%s")
```
jbedard, line 40: Nothing replaced this?

jbedard (approved):

lint/ruff.bzl
alexeagle: Right, instead of explicitly telling it which config to use, we let it look among the configs it can find.

honnix: I'm not sure this is a correct approach. I think ruff will only be able to find config file in project root, which may not always be the case.

PR #62, opened 2023-12-01
feat: generate a patch file with linter fixes

### LINKED ISSUE #18: [FR]: Apply fixes

dzbarsky:
Generate a patch file with linter fixes. The patch may then be applied by whatever tool is running Bazel.

[asciicast](https://asciinema.org/a/r9JKJ8uKgAZTzlUPdDdHlY1CB)

---

### Type of change

### Test plan

- New `bats` test cases added

```
./lint.sh --fix //src:all
```

and with debug flag:

```
./lint.sh //src:ts --@aspect_rules_lint//lint:debug
```

Fixes #18


lint/private/patcher.mjs
```diff
@@ -0,0 +1,55 @@
...
+    process.exit(1);
+}
+
+const diffOut = fs.createWriteStream(process.env["PATCH_PATH"]);
+
+for (let i = 0; i < filesToLint.length; i++) {
+    const origF = filesToLint[i];
+    const newF = tmpFiles[i];
+    console.log(origF, newF);
+    const results = childProcess.spawnSync("diff", ["-U8", origF, newF], {
```
alexeagle, line 46: I think we can't rely on tools being on the PATH, and being portable mac/linux. Does coreutils provide `diff`?

lint/private/patcher.mjs
```diff
@@ -0,0 +1,55 @@
...
+    fs.chmodSync(destF, "600");
+    tmpFiles.push(destF);
+}
+
+const filesIndex = linterArgs.indexOf("$$FILES");
+linterArgs.splice(filesIndex, 1, ...tmpFiles);
+
+const linterPath = path.join(process.env["JS_BINARY__EXECROOT"], process.env["LINTER_PATH"]);
+const newEnv = {...process.env};
+delete newEnv["JS_BINARY__EXIT_CODE_OUTPUT_FILE"];
```
alexeagle, line 28: I think you won't need this anymore after rebase

lint/private/patcher.mjs
```diff
@@ -0,0 +1,55 @@
+import childProcess from 'child_process';
```
alexeagle, line 1: needs some doc like "wrapper around linter tools to create a writeable sandbox of source files, then collect the resulting edits as a patch"

gregmagolan:
Related to https://github.com/aspect-build/rules_lint/issues/18

lint/private/patcher.mjs
dzbarsky: Nope, its https://www.gnu.org/software/diffutils/

gregmagolan:
@alexeagle and I paired on this today and moved the implementation along. Works for eslint.

Things left todo:
- wire the patcher up for the other linters
- improve test to check that the contents of the generated patch files are asserted

While working on this, a rules_lint bug came to mind that should be address in a separate PR that filters out generated srcs from the list of files to lint for all linters so that lint warnings & failures don't created for generated files. For example a `ts_project` target with `.ts` files in srcs that are outputs of other targets.

lint/private/patcher.mjs
```diff
@@ -0,0 +1,55 @@
...
+    fs.mkdirSync(path.dirname(destF), {recursive: true});
+    fs.cpSync(f, destF);
+    fs.chmodSync(destF, "600");
+    tmpFiles.push(destF);
+}
+
+const filesIndex = linterArgs.indexOf("$$FILES");
+linterArgs.splice(filesIndex, 1, ...tmpFiles);
+
+const linterPath = path.join(process.env["JS_BINARY__EXECROOT"], process.env["LINTER_PATH"]);
```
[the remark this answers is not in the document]
gregmagolan: Fixed in the updated implementation

lint/private/patcher.mjs
gregmagolan: Addressed in the updated implementation.

lint/private/patcher.mjs
alexeagle: I'll add a TODO for this

alexeagle:
> don't rely on the `diff` tool being on the path

I left behind a TODO for that one.

The other AIs are complete now, and I've updated the docs. I think it's ready to merge.

alexeagle (approved):
🌮 🐈

lint/private/patcher.mjs
alexeagle: TODONE by using the diff from cosmopolitan (provided by the "ape" bazel module)

PR #70, opened 2023-12-05
fix: Respect exclude when formatting with ruff

jsharpe:
As per discussion here: https://github.com/astral-sh/ruff/discussions/5857#discussioncomment-6583943 pass `--force-exclude` when invoking `ruff format`

---

### Type of change

ruff format now respects exclude and extend-exclude directives in ruff.toml configuration files.

### Test plan

Add a directory path to an extend-exclude directive in ruff.toml and observe that the files in that directory are not formatted.

### DIFF
--- a/format/private/format.sh
+++ b/format/private/format.sh
@@ -62,7 +62,7 @@ case "$mode" in
  fix)
    swiftmode=""
    prettiermode="--write"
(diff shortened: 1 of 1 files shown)


alexeagle (approved):
Thanks!

PR #71, opened 2023-12-06
chore: change maintainer to individual rather than team.

alexeagle:
I saw us making this change in other repos so it should be consistent. Maybe it also explains why I didn't get an email about rules_lint 0.6.1 not publishing a PR to BCR


gregmagolan (approved):
🦖

PR #78, opened 2023-12-11
feat: add mypy

alexeagle:
Demo:

```
$ ./lint.sh src:unused_import
INFO: Analyzed target //src:unused_import (1 packages loaded, 470 targets configured).
ERROR: /home/alexeagle/Projects/rules_lint/example/src/BUILD.bazel:21:11: mypy src/mypy.unused_import.aspect_rules_lint.report failed: (Exit 1): mypy failed: error executing mypy command (from target //src:unused_import) bazel-out/k8-opt-exec-ST-13d3ddad9198/bin/tools/mypy src/unused_import.py

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
src/unused_import.py:18: error: Incompatible types in assignment (expression has type "float", variable has type "int")  [assignment]
Found 1 error in 1 file (checked 1 source file)
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 1.101s, Critical Path: 0.88s
INFO: 2 processes: 2 internal.
ERROR: Build did NOT complete successfully
```

### Type of change

### Test plan

- New test cases added


lint/mypy.bzl
```diff
@@ -0,0 +1,190 @@
...
+with appreciation to [Jonathon Belotti](https://github.com/thundergolfer)
+and the other [contributors](https://github.com/bazel-contrib/bazel-mypy-integration/graphs/contributors)
+especially [David Zbarsky](https://github.com/dzbarsky)
+
+### TODO
+
+- Document --norun_validations
+- Support third-party definitions from stubs installed from type-* packages
+- Allow configured typeshed repo, e.g. args.add("--custom-typeshed-dir", "external/my_typeshed")
+- Set MYPYPATH if needed for cases where 1p sources don't resolve correctly to transitive .py sources: https://github.com/dzbarsky/mypy-bazel-experiments/blob/master/build/py/mypy.bzl#L163
```
DavidZbarsky-at, line 24: fyi this repo isn't public, but if this is just notes for you that's fine :)

lint/mypy.bzl
```diff
@@ -0,0 +1,190 @@
...
+        transitive_cache_map = depset(
+            direct = direct,
+            transitive = [d[MypyInfo].transitive_cache_map for d in deps if MypyInfo in d],
+        ),
+    )
+
+def _extract_transitive_inputs(deps):
+    result = []
+    for dep in deps:
+        # TODO: look through runfiles of 3p packages where .pyi files would appear in data.
```
DavidZbarsky-at, line 48: I remember we (me/Matt maybe?) were chatting a bit about surfacing these in the PyInfo provider, if that's not tractable maybe rules_py could hack something in? I suspect it would be both cleaner and better for perf than trying to find them here

lint/mypy.bzl
```diff
@@ -0,0 +1,190 @@
...
+
+        # NB: handle //path/to/package:subdir/file.py -> subdir/file.py
+        path = paths.relativize(src.path, ctx.label.package)
+        for kind in ("meta", "data"):
+            # Note, this assumes every source file has a single mypy aspect producing these output files
+            # For subdir/file.py, declare subdir/file.meta.json
+            file = ctx.actions.declare_file(paths.replace_extension(path, ".%s.json" % kind))
+            cache_map_outputs.append(file)
+            cache_map.append(file)
+    args.add_all(["--cache-map"] + cache_map)
```
DavidZbarsky-at, line 121: nit: doesn't this copy the list? could do `args.add('--cache-map'); args.add_all(cache_map)` instead

lint/mypy.bzl
alexeagle: Yeah once this gets more mature I'll look at boiling up some of the parts into proposals on rules_python

lint/mypy.bzl
alexeagle: yeah it's meant for my own notes, I could move out of the block that's docgen'ed

lint/mypy.bzl
alexeagle: left a TODO for upstreaming something about stubs to rules_python.

lint/mypy.bzl
```diff
@@ -0,0 +1,231 @@
...
+```
+# Add mypy type-check validation actions to py_library targets.
+# Disable for a particular build with --norun_validations
+build --aspects=//tools:lint.bzl%mypy
+```
+
+### Reporting
+
+Unlike most linters hosted in rules_lint, mypy produces only error semantics.
+That means that typecheck violations will input_depsets in failed build actions, rather than
```
mattem, line 25: I'm not sure I can make sense of this statement, "will input_depsets"?

lint/mypy.bzl
```diff
@@ -0,0 +1,231 @@
...
+def _extract_transitive_inputs(deps):
+    input_depsets = []
+    for dep in deps:
+        # Supply the --cache-map files from transitives
+        if MypyInfo in dep:
+            input_depsets.append(dep[MypyInfo].transitive_cache_map)
+
+        # NB: relies on PyInfo being a Bazel global symbol, one day we should be forced to load it
+        # TODO: maybe we can avoid passing .py source files when the cache-map files were found?
+        if PyInfo in dep:
```
mattem, line 71: Should load through rules_python now as `incompatible_python_disallow_native_rules` might be enabled

lint/mypy.bzl
```diff
@@ -0,0 +1,231 @@
...
+
+    # Without this option, mypy will print on success, which will spam the Bazel output:
+    # Success: no issues found in 1 source file
+    args.add("--no-error-summary")
+
+    # Enable with --aspects_parameters=rules_lint_verbose=True
+    if ctx.attr.rules_lint_verbose:
+        args.add("--verbose")
+
+    # TODO: need some of these? Seen at RH
```
mattem, line 172: Think this should be removed.

lint/mypy.bzl
```diff
@@ -0,0 +1,231 @@
...
+
+    ctx.actions.run(
+        inputs = inputs,
+        outputs = cache_map_outputs,
+        executable = executable,
+        arguments = [args],
+        env = {
+            # TODO: investigate a more robust way to construct this, and avoid the length of the
+            # value overflowing the operating system limit.
+            "PYTHONPATH": ":".join(["external/" + t for t in transitive_imports.to_list()]),
```
mattem, line 188: Does this clobber over the binaries own `PYTHONPATH`? What happens when mypy has plugins?

lint/mypy.bzl
alexeagle: oops, was an over-eager searchreplace

lint/mypy.bzl
alexeagle: I agree, though I was trying to avoid the rules_python dependency. I think it's needed for provider symbols which is why this repo depends on rules_js too.

alexeagle: Actually I can't enable the flag, seems like an upstream problem. Filed https://github.com/bazelbuild/rules_python/issues/1645

alexeagle: added a TODO

lint/mypy.bzl
alexeagle: No idea about plugins, will add to the TODO list.

As for how the rules_python runtime stub handles it, it seems to merge the entries:

```
old_python_path = os.environ.get('PYTHONPATH')
  if old_python_path:
    python_path_entries += old_python_path.split(os.pathsep)
```

alexeagle: Anyhow I'd like to delete this line altogether if we can make a venv to run inside instead.

DavidZbarsky-at: the other relevant var I saw was MYPYPATH, I forget the difference at this point but maybe that's useful to avoid confusion/clobbering here?

DavidZbarsky-at (approved):

lint/mypy.bzl
alexeagle: @mattem tracked that down - files on the MYPYPATH will be type-checked, in this case we just want the file available for the checker to locate it but want to avoid type-checking 3p code

alexeagle:
not planning to land this, per #85

Issue #88, opened 2023-12-22
[Bug]: Format fails when there are deleted files


psalaberria002:
### What happened?

The format script uses git ls-files, which includes deleted files if the changes aren't committed.

If we used the following command deleted files would be excluded from the output.
`$ git ls-files "*.bazel" | grep -vE "^$(git ls-files -d | paste -sd "|" -)$"`

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: aspect 5.8.5

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:

Language(s) and/or frameworks involved:

### How to reproduce

```shell
$ git clone git@github.com:aspect-build/rules_lint.git
$ cd rules_lint/example
$ mv src/hello.jsonnet src/hello2.jsonnet
$ bazel run format
Opening input file: src/hello.jsonnet: no such file or directory
```

PR #89, opened 2023-12-22
Exclude deleted files and include untracked ones from format

psalaberria002:
With this approach deleted files are excluded, and untracked files are included.

Fixes https://github.com/aspect-build/rules_lint/issues/88

### DIFF
--- a/format/private/format.sh
+++ b/format/private/format.sh
@@ -79,7 +79,7 @@ case "$mode" in
 esac
 
 if [ "$#" -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


alexeagle (approved):
Thanks - do any of these new flags have a long form? That makes it more self-documenting

psalaberria002:
> Thanks - do any of these new flags have a long form? That makes it more self-documenting

They do. I have added them now.

alexeagle:
Thank you!!

PR #91, opened 2024-01-03
Ruff Windows Support

LaurenceTews:
* Adds Windows support for Ruff. I understand this repository doesn't support Windows as a whole yet, but this may prove useful to folks not using the CLI yet. 

### Type of change

### Test plan

Changes verified by running `bazel run //tools:ruff check` on Windows x64 machine, with appropriate alias configured.


alexeagle (approved):
thanks!

PR #95, opened 2024-01-05
Add golangci-lint as an aspect

### LINKED ISSUE #82: [FR]: support golangci-lint

No support for Go linting

### Describe the feature

Probably the most popular/common linter for Go is https://golangci-lint.run

It would be wonderful to support this.
Thanks for the awesome project :)

psalaberria002:
### Type of change

Fixes #82 

### Test plan

- New test cases added
- Manual testing; please provide instructions so we can reproduce: `(cd example && ./lint.sh //...)`

### Notes
Golang-ci lint relies on the go binary being in the PATH. I had to add rules_go as a dependency in the main repo for passing the toolchain type. I am not sure we can do this without that dependency.

### DIFF
--- a/example/MODULE.bazel
+++ b/example/MODULE.bazel
@@ -8,7 +8,7 @@ bazel_dep(name = "rules_buf", version = "0.2.0")
 bazel_dep(name = "bazel_skylib", version = "1.4.2")
 bazel_dep(name = "rules_java", version = "5.5.0")
 bazel_dep(name = "rules_jvm_external", version = "4.5")
-bazel_dep(name = "rules_go", version = "0.42.0")
+bazel_dep(name = "rules_go", version = "0.42.0", repo_name = "io_bazel_rules_go")
 bazel_dep(name = "rules_proto", version = "5.3.0-21.7")
 bazel_dep(name = "rules_python", version = "0.26.0")
 bazel_dep(name = "buildifier_prebuilt", version = "6.3.3")
@@ -50,7 +50,7 @@ pip.parse(
 )
 use_repo(pip, "pip")
 
-go_sdk = use_extension("@rules_go//go:extensions.bzl", "go_sdk")
+go_sdk = use_extension("@io_bazel_rules_go//go:extensions.bzl", "go_sdk")
 go_sdk.download(
     name = "go_sdk",
     version = "1.20.3",
--- a/example/src/hello.go
+++ b/example/src/hello.go
@@ -9,5 +9,6 @@ const (
 )
 
 func main() {
-	fmt.Printf("Hello %s\n", w)
+	hello := fmt.Sprintf("Hello %s\n", w)
+	fmt.Printf(hello)
 }
--- a/example/src/hello.go
+++ b/example/src/hello.go
@@ -1,6 +1,8 @@
 package main
 
-import "fmt"
+import (
+	"fmt"
+)
 
 const (
 	w = "world"
--- a/lint/golangci-lint.bzl
+++ b/lint/golangci-lint.bzl
@@ -10,7 +10,6 @@ golangci_lint = golangci_lint_aspect(
 ```
 """
 
-load("@bazel_skylib//rules:native_binary.bzl", "native_binary")
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@io_bazel_rules_go//go:def.bzl", "go_context")
 load("//lint/private:lint_aspect.bzl", "report_file")
@@ -95,23 +94,6 @@ def golangci_lint_aspect(binary, config):
             ),
         },
         toolchains = ["@io_bazel_rules_go//go:toolchain"],
-    )
-
-def golangci_lint_binary(name):
-    """Wrapper around native_binary to select the correct golangci-lint executable for the execution platform."""
-    native_binary(
-        name = name,
-        src = select(
-            {
-                "@bazel_tools//src/conditions:linux_x86_64": "@golangci_lint_linux_x86_64//:golangci-lint",
-                "@bazel_tools//src/conditions:linux_aarch64": "@golangci_lint_linux_aarch64//:golangci-lint",
-                "@bazel_tools//src/conditions:darwin_x86_64": "@golangci_lint_macos_x86_64//:golangci-lint",
-                "@bazel_tools//src/conditions:darwin_arm64": "@golangci_lint_macos_aarch64//:golangci-lint",
-            },
-            no_match_error = "golangci-lint hasn't been fetched for your platform",
-        ),
-        out = "golangci-lint",
-        visibility = ["//visibility:public"],
     )
 
 # buildifier: disable=function-docstring
--- a/lint/golangci-lint.bzl
+++ b/lint/golangci-lint.bzl
@@ -96,33 +96,39 @@ def golangci_lint_aspect(binary, config):
         toolchains = ["@io_bazel_rules_go//go:toolchain"],
     )
 
-# buildifier: disable=function-docstring
-def fetch_golangci_lint():
+def fetch_golangci_lint(version = "1.55.2"):
+    """Naive macro that fetches a specific version of the golangci-lint from GitHub releases, for commonly-used platforms
+
+    Args:
+        version: must be the default value. In the future this could be honored when we support multiple versions.
+    """
+    if version != "1.55.2":
+        fail("Only a single version of golangci-lint is currently mirrored. Please file an issue if you need a different version.")
     http_archive(
         name = "golangci_lint_linux_x86_64",
         build_file_content = "exports_files([\"golangci-lint\"])",
-        strip_prefix = "golangci-lint-1.55.2-linux-amd64",
+        strip_prefix = "golangci-lint-{}-linux-amd64".format(version),
         sha256 = "ca21c961a33be3bc15e4292dc40c98c8dcc5463a7b6768a3afc123761630c09c",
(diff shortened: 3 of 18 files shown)


lint/golangci-lint.bzl
```diff
@@ -0,0 +1,126 @@
...
+golangci_lint = golangci_lint_aspect(
+    binary = "@@//tools:golangci-lint",
+    config = "@@//:.golangci.yaml",
+)
+```
+"""
+
+load("@bazel_skylib//rules:native_binary.bzl", "native_binary")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load("@rules_go//go:def.bzl", "go_context")
```
alexeagle, line 15: Yeah it's unfortunate that this module ends up with dependencies on every language, but it already happened with rules_js and I expect providers will force us to do it more.

Now that bzlmod is enabled by default in the LTS Bazel version, I'm less concerned about the impacts on users.

alexeagle:
Thanks! To get it green, you'll have to add a rules_go setup to the example/WORKSPACE.bazel file too, since we are testing that it works with bzlmod disabled.

lint/golangci-lint.bzl
```diff
@@ -0,0 +1,126 @@
...
+    )
+
+def golangci_lint_binary(name):
+    """Wrapper around native_binary to select the correct golangci-lint executable for the execution platform."""
+    native_binary(
+        name = name,
+        src = select(
+            {
+                "@bazel_tools//src/conditions:linux_x86_64": "@golangci_lint_linux_x86_64//:golangci-lint",
+                # TODO: Add other platform binaries
```
alexeagle, line 107: I think we need to add these before landing, I can help if you need :)

psalaberria002: Done.

example/MODULE.bazel
```diff
@@ -8,7 +8,7 @@ bazel_dep(name = "rules_buf", version = "0.2.0")
 bazel_dep(name = "bazel_skylib", version = "1.4.2")
 bazel_dep(name = "rules_java", version = "5.5.0")
 bazel_dep(name = "rules_jvm_external", version = "4.5")
-bazel_dep(name = "rules_go", version = "0.42.0")
+bazel_dep(name = "rules_go", version = "0.42.0", repo_name = "io_bazel_rules_go")
```
alexeagle, line 11: I guess this alias name is unavoidable for some reason?

lint/golangci-lint.bzl
```diff
@@ -0,0 +1,146 @@
...
+                default = config,
+                allow_single_file = True,
+            ),
+        },
+        toolchains = ["@io_bazel_rules_go//go:toolchain"],
+    )
+
+def golangci_lint_binary(name):
+    """Wrapper around native_binary to select the correct golangci-lint executable for the execution platform."""
+    native_binary(
```
alexeagle, line 102: could you move this into the example folder to match the others? we haven't been opinionated about the select statement (possible that some users have unique conditions)

lint/golangci-lint.bzl
```diff
@@ -0,0 +1,146 @@
...
+        strip_prefix = "golangci-lint-1.55.2-linux-amd64",
+        sha256 = "ca21c961a33be3bc15e4292dc40c98c8dcc5463a7b6768a3afc123761630c09c",
+        urls = ["https://github.com/golangci/golangci-lint/releases/download/v1.55.2/golangci-lint-1.55.2-linux-amd64.tar.gz"],
+    )
+    http_archive(
+        name = "golangci_lint_linux_aarch64",
+        build_file_content = "exports_files([\"golangci-lint\"])",
+        strip_prefix = "golangci-lint-1.55.2-linux-arm64",
+        sha256 = "8eb0cee9b1dbf0eaa49871798c7f8a5b35f2960c52d776a5f31eb7d886b92746",
+        urls = ["https://github.com/golangci/golangci-lint/releases/download/v1.55.2/golangci-lint-1.55.2-linux-arm64.tar.gz"],
```
alexeagle, line 131: how often does it do releases? I suspect we'll need an update automation like we have for `ruff` but it can be a later PR.

lint/golangci-lint.bzl
```diff
@@ -0,0 +1,146 @@
...
+    native_binary(
+        name = name,
+        src = select(
+            {
+                "@bazel_tools//src/conditions:linux_x86_64": "@golangci_lint_linux_x86_64//:golangci-lint",
+                "@bazel_tools//src/conditions:linux_aarch64": "@golangci_lint_linux_aarch64//:golangci-lint",
+                "@bazel_tools//src/conditions:darwin_x86_64": "@golangci_lint_macos_x86_64//:golangci-lint",
+                "@bazel_tools//src/conditions:darwin_arm64": "@golangci_lint_macos_aarch64//:golangci-lint",
+            },
+            no_match_error = "golangci-lint hasn't been fetched for your platform",
```
alexeagle, line 111: could you also say what the user ought to do about it? (file a bug, etc)

lint/golangci-lint.bzl
```diff
@@ -0,0 +1,146 @@
...
+        ctx: Bazel Rule or Aspect evaluation context
+        executable: label of the the golangci-lint program
+        srcs: golang files to be linted
+        config: label of the .golangci.yaml file
+        report: output file to generate
+        use_exit_code: whether to fail the build when a lint violation is reported
+    """
+
+    # golangci-lint calls out to Go, so we need the go context
+    go = go_context(ctx)
```
alexeagle, line 35: Any other link to docs in rules_go about this?

lint/golangci-lint.bzl
```diff
@@ -0,0 +1,146 @@
...
+    inputs = srcs + [config] + go.sdk_files
+    args = ctx.actions.args()
+    args.add_all(srcs)
+
+    command = """#!/usr/bin/env bash
+        export GOROOT=$(cd "$(dirname {go_tool})/.."; pwd)
```
alexeagle, line 41: would be nice to link to docs here as well, just in case rules_go changes something that requires users to account for it like adding another variable

alexeagle (commented):
cool, few small things

example/MODULE.bazel
psalaberria002: I was getting errors, yes. Not sure it is avoidable.

lint/golangci-lint.bzl
psalaberria002: Now that I am moving it to the example folder I am gonna get rid of it. It doesn't add much value, and the solution isn't always the same.

lint/golangci-lint.bzl
psalaberria002: every 1-3 months. Is renovate configured in the repo? If so, it should handle the http_archives in a batch.

lint/golangci-lint.bzl
psalaberria002: I took this from our internal repo. Probably based on some discussions in the Bazel Slack workspace. Can't find any refs in the docs.

alexeagle: okay, thanks for looking.

lint/golangci-lint.bzl
alexeagle: Hm, yeah renovate is one possibility. In the ruff case you can see we mirror a list of versions so that users can pin. I can pick this up either way, thanks.

example/MODULE.bazel
alexeagle: okay I'll take a stab at removing it in a followup, maybe. It does feel like the name we choose here leaks into userspace, at least for WORKSPACE users.

psalaberria002:
Thanks for polishing the PR 😉

example/src/hello.go
```diff
@@ -1,6 +1,8 @@
 package main
 
-import "fmt"
+import (
```
alexeagle, line 3: this file needs a lint violation so I can see that it's actually working. Can you suggest a short one?

psalaberria002: Added one.
```
INFO: From golangcilint src/golangcilint.hello_go.aspect_rules_lint.report:
../../bazel-source-roots/0/src/hello.go:13:2: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)
        fmt.Printf(hello)
```

lint/golangci-lint.bzl
```diff
@@ -0,0 +1,134 @@
...
+    Args:
+        version: must be the default value. In the future this could be honored when we support multiple versions.
+    """
+    if version != "1.55.2":
+        fail("Only a single version of golangci-lint is currently mirrored. Please file an issue if you need a different version.")
+    http_archive(
+        name = "golangci_lint_linux_x86_64",
+        build_file_content = "exports_files([\"golangci-lint\"])",
+        strip_prefix = "golangci-lint-{}-linux-amd64".format(version),
+        sha256 = "ca21c961a33be3bc15e4292dc40c98c8dcc5463a7b6768a3afc123761630c09c",
```
psalaberria002, line 111: Isn't this problematic? If a user provides a different version as an argument to the fetch_golangci_lint function the sha256 will be wrong.

alexeagle: yes, that's why it only allows the one version we currently support. We'll have to refactor this to a dictionary of hashes per release at some point.

example/src/hello.go
alexeagle: cool thanks

alexeagle (approved):

PR #96, opened 2024-01-09
fix: filter out generated files from linting

### LINKED ISSUE #90: [Bug]: Lints reported on generated files
### What happened?

Haven't verified, but we think we currently run linters on all `srcs` of a `*_library` including those in `bazel-out` which means we're reporting lint violations on generated code. This is not desirable, we should probably filter them out.

Follow-up from #62 

### Version

HEAD

### How to reproduce

_No response_

### Any other information?

_No response_

alexeagle:
Fixes #90

---

### Type of change

### Test plan

- New test cases added

### DIFF
--- a/lint/ruff.bzl
+++ b/lint/ruff.bzl
@@ -90,7 +90,7 @@ def ruff_fix(ctx, executable, srcs, config, patch):
         output = patch_cfg,
         content = json.encode({
             "linter": executable._ruff.path,
-            "args": ["check", "--fix"] + [s.short_path for s in srcs],
+            "args": ["check", "--fix"] + [s.path for s in srcs],
             "files_to_diff": [s.path for s in srcs],
             "output": patch.path,
         }),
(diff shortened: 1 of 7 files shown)


lint/ruff.bzl
```diff
@@ -90,7 +90,7 @@ def ruff_fix(ctx, executable, srcs, config, patch):
         output = patch_cfg,
         content = json.encode({
             "linter": executable._ruff.path,
-            "args": ["check", "--fix"] + [s.short_path for s in srcs],
+            "args": ["check", "--fix"] + [s.path for s in srcs],
```
jbedard, line 93: Why was this needed?

alexeagle: Greg asked for it on #90 :)
It's unrelated.

gregmagolan (approved):
🌮

alexeagle:
Hm, yeah that's tricky. And I appreciate that this change happened in the same release that includes your PRs so you can't even use them yet! :)

A few ways I could see:
1. Don't bother with this exclusion, revert this PR, and rather document how you should configure each tool (e.g. shellcheckrc) to exclude files in `bazel-out/*` for typical use cases
2. A global aspect_attribute like fail_on_violation - but I suspect that's too coarse, most generated files shouldn't be linted even if some are
3. Find a way for the aspect to visit your transform target instead of the `sh_binary` that receives the files. But, your non-injected file might not even be valid bash, if the placeholders use special syntax. 
4. Use something else hung off the dependency graph (likely `tags=["something"]`) so that when the aspect visits a target we can decide if the srcs should be filtered with `is_source`

gregmagolan:
Not linting generated files is the 99% percent case. I like the idea of a using a tag on the target to opt-in to linting generated files. The suggested fixes with the injections, however, won't line up line-for-line with the original sources in that case.

alexeagle:
@psalaberria002 would that work for you if you have to mark the targets produced by your macro with something like `tags=["lintable"]` to hint to us that the `is_source` filter should be disabled when visiting that target?

And yeah @gregmagolan that's true that the fixer should not honor this, and should continue to visit only the source files.

psalaberria002:
That would work yes. Probably that's the simplest approach.

PR #98, opened 2024-01-11
Enable shellcheck aspect for sh_binary

psalaberria002:
### Type of change

If the user has srcs in an sh_binary with shellcheck violations then it will a breaking change for them.
I don't think it deserves a major version bump tho.

### Test plan

- Manual testing; please provide instructions so we can reproduce: Add an sh_binary and see that their srcs are linted.


alexeagle (approved):
thanks

PR #101, opened 2024-01-16
feat: Add clang-format for formatting C/C++

jsharpe:
---

### Type of change

Add support for formatting C/C++ code using clang-format.

### Test plan

- New test cases added
Added example C++ file to show it is formatted by clang-format when format is run.

### DIFF
--- a/example/tools/BUILD
+++ b/example/tools/BUILD
@@ -131,7 +131,7 @@ shellcheck_binary(name = "shellcheck")
 
 multi_formatter_binary(
     name = "format",
-    clangformat = "@llvm_toolchain_llvm//:bin/clang-format",
+    cc = "@llvm_toolchain_llvm//:bin/clang-format",
     go = "@go_sdk//:bin/gofmt",
     java = ":java-format",
     javascript = ":prettier",
--- a/example/tools/BUILD
+++ b/example/tools/BUILD
@@ -131,6 +131,7 @@ shellcheck_binary(name = "shellcheck")
 
 multi_formatter_binary(
     name = "format",
+    clangformat = "@llvm_toolchain_llvm//:bin/clang-format",
     go = "@go_sdk//:bin/gofmt",
     java = ":java-format",
     javascript = ":prettier",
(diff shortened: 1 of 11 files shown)


example/tools/BUILD
```diff
@@ -131,6 +131,7 @@ shellcheck_binary(name = "shellcheck")
 
 multi_formatter_binary(
     name = "format",
+    clangformat = "@llvm_toolchain_llvm//:bin/clang-format",
```
alexeagle, line 134: ah, that's so convenient that the toolchain already includes the format binary 😌

format/private/format.sh
```diff
@@ -212,6 +214,17 @@ if [ -n "$files" ] && [ -n "$bin" ]; then
   fi
 fi
 
+if [ "$#" -eq 0 ]; then
+  files=$(git ls-files --cached --modified --other --exclude-standard '*.cc' '*.cpp' '*.cxx' '*.c++' '*.C' '*.c' '*.h' '*.hh' '*.hpp' '*.ipp' '*.hxx' '*.h++' '*.inc' '*.inl' '*.tlh' '*.tli' '*.H' | { grep -vE "^$(git ls-files --deleted)$" || true; } )
```
alexeagle, line 218: sorry, I know this file could really use some refactoring...

format/private/formatter_binary.bzl
```diff
@@ -18,6 +18,7 @@ _TOOLS = {
     "sql": "prettier-sql",
     "sh": "shfmt",
     "protobuf": "buf",
+    "clangformat": "clang-format",
```
alexeagle, line 21: in other cases we tried to have the user-visible part (the name of the attribute, so the key in this dict) be the name of the language, rather than the name of a tool. Should we make it `cpp` or is that confusing?

alexeagle (commented):
thanks!

format/private/formatter_binary.bzl
jsharpe: Yes or `cc`?

example/tools/BUILD
jsharpe: It also includes `clang-tidy` so paves the way for including linting too.

format/private/formatter_binary.bzl
jsharpe: I went with `cc` as this matches `rules_cc`

alexeagle (approved):

PR #102, opened 2024-01-16
Fix macos

alexeagle:
Also includes an unrelated bump to latest aspect CLI for the `lint --fix` command, and Bazel 7.0 final


example/MODULE.bazel
```diff
@@ -27,7 +27,10 @@ use_repo(rules_ts_ext, "npm_typescript")
 # Configure and register the toolchain.
 llvm = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm")
 llvm.toolchain(
-    llvm_version = "16.0.0",
+    llvm_versions = {
+        "": "16.0.0",
+        "darwin-x86_64": "15.0.7",
```
jbedard, line 32: Why is the downgrade needed for macos? Why not downgrade for every os?

jbedard (approved):

example/MODULE.bazel
alexeagle: they just don't publish LLVM for all platforms at each release, for some reason,

https://github.com/grailbio/bazel-toolchain/tree/master?tab=readme-ov-file#per-host-architecture-llvm-version

PR #103, opened 2024-01-17
Vale

alexeagle:
Add a linter for prose in markdown files, see https://vale.sh/

---

### Type of change

### Test plan

- New test cases added


alexeagle:
@mattem since our linting is based on visiting targets, we need users to declare their Markdown files in some library rule.
@dwtj did something like that in https://github.com/dwtj/dwtj_rules_markdown - not sure if we want to introduce something similar.

For now I've used a `filegroup` with `tags = ["markdown"]` which I think is *maybe* a viable option?

mattem (approved):

format/repositories.bzl
```diff
@@ -126,6 +126,7 @@ def fetch_java_format():
 def fetch_ktfmt():
     http_jar(
         name = "ktfmt",
+        integrity = "sha256-l/x/vRlNAan6RdgUfAVSQDAD1VusSridhNe7TV4/SN4=",
```
alexeagle, line 129: sorry, unrelated fix I noticed while testing

lint/vale.bzl
```diff
@@ -0,0 +1,184 @@
...
+                tag = tag,
+                plat = plat,
+                version = version,
+                ext = "zip" if is_windows else "tar.gz",
+            ),
+            sha256 = sha256,
+            build_file_content = """exports_files(["vale", "vale.exe"])""",
+        )
+
+        fetch_styles()
```
alexeagle, line 184: note, I chose to do this automatically, but it's also reasonable to require users to call both `fetch_vale` and `fetch_vale_styles`

lint/vale_library.bzl
```diff
@@ -0,0 +1,112 @@
...
+        "homepage": "https://redhat-documentation.github.io/vale-at-red-hat/docs/main/user-guide/openshift-asciidoc-style-for-vale/",
+        "integrity": "sha256-F6LMXFKfHIaFhG26BazlN3W2m1DN628f0YTe+An1TSQ=",
+        "version": "v431",
+        "url": "https://github.com/redhat-documentation/vale-at-red-hat/releases/download/{}/OpenShiftAsciiDoc.zip",
+    },
+]
+
+VALE_STYLES = [s["name"] for s in VALE_STYLE_DATA]
+
+def fetch_styles():
```
alexeagle, line 103: Note: this api could accept an optional list of which styles to fetch, but I don't think it matters since unreferenced ones won't be fetched anyway.

PR #105, opened 2024-01-18
refactor: overhaul of format.sh

alexeagle:
Use GitHub Linguist to define the file extensions for each language, getting much more accurate. Extract a helper function for listing files we want to format to reduce duplication.

---

### Type of change

### Test plan

Ran format.sh after changing code in each language


format/private/mirror_linguist_languages.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+# Generate a snippet of bash code to help us detect languages based on filenames
+raw=$(mktemp --suffix=.yml)
+json=$(mktemp --suffix=.json)
+wget -O "$raw" https://raw.githubusercontent.com/github-linguist/linguist/master/lib/linguist/languages.yml
+yq -o=json '.' "$raw" > "$json"
```
alexeagle, line 6: I'm sure YQ could do this in one step, but I'm already familiar with JQ and it wasn't the right day to learn a new tool.

thesayyn: we don't need yq anymore. we can continue using jq

alexeagle: the input is a yaml file though, so I think I do need this conversion?

Obviously this code will barely ever be run, and one could just look at the github-linguist repo and see a short list of extensions for the next language added. Maybe it's overkill to check this in at all?

format/private/format.sh
```diff
@@ -19,32 +22,72 @@ function on_exit {
...
+      'Protocol Buffer') patterns=('*.proto') ;;
+      'Python') patterns=('.gclient' 'DEPS' 'SConscript' 'SConstruct' 'wscript' '*.py' '*.cgi' '*.fcgi' '*.gyp' '*.gypi' '*.lmi' '*.py3' '*.pyde' '*.pyi' '*.pyp' '*.pyt' '*.pyw' '*.rpy' '*.spec' '*.tac' '*.wsgi' '*.xpy') ;;
+      'SQL') patterns=('*.sql' '*.cql' '*.ddl' '*.inc' '*.mysql' '*.prc' '*.tab' '*.udf' '*.viw') ;;
+      'Scala') patterns=('*.scala' '*.kojo' '*.sbt' '*.sc') ;;
+      'Shell') patterns=('.bash_aliases' '.bash_functions' '.bash_history' '.bash_logout' '.bash_profile' '.bashrc' '.cshrc' '.flaskenv' '.kshrc' '.login' '.profile' '.zlogin' '.zlogout' '.zprofile' '.zshenv' '.zshrc' '9fs' 'PKGBUILD' 'bash_aliases' 'bash_logout' 'bash_profile' 'bashrc' 'cshrc' 'gradlew' 'kshrc' 'login' 'man' 'profile' 'zlogin' 'zlogout' 'zprofile' 'zshenv' 'zshrc' '*.sh' '*.bash' '*.bats' '*.cgi' '*.command' '*.fcgi' '*.ksh' '*.sh.in' '*.tmux' '*.tool' '*.trigger' '*.zsh' '*.zsh-theme') ;;
+      'Starlark') patterns=('BUCK' 'BUILD' 'BUILD.bazel' 'MODULE.bazel' 'Tiltfile' 'WORKSPACE' 'WORKSPACE.bazel' '*.bzl' '*.star') ;;
```
thesayyn, line 48: hmm, BUCK files might contain types. does our formatter understand that?

alexeagle: This is a problem generally, we use a tool that "formats code in some language" but there are syntactic variants for all of them that we can't possibly be experts on.
I think maybe the right answer (in a follow-up PR?) is to provide an affordance for this list of patterns to be overridden from the `multi_formatter_binary` rule

format/private/mirror_linguist_languages.sh
thesayyn: oh missed that the extension was yaml.

thesayyn (approved):

PR #110, opened 2024-01-20
feat: escape hatch from is_source filtering

alexeagle:
---

### Type of change

### Test plan

- New test cases added
well, it's a tags=manual test. I should make lint_test have an expected_exit_code.


alexeagle:
Followup to #96 FYI @psalaberria002

gregmagolan (approved):
🌮

PR #114, opened 2024-01-22
Add gofumpt as an alternative go formatter

jelmer:
gofumpt is a popular alternative formatter for Go that is compatible with gofmt but stricter

This is a new feature, mainly adding the requisite fetch_* and alias for gofumpt and providing a suggestion in the example BUILD.bazel file.

This code was tested by changing the go formatter in the example to ":gofumpt" and running //:format.

Suggested release note: "Add supporting code for using the gofumpt formatter for Go".


alexeagle (commented):
Hey @jelmer thanks for contributing!

Do you know whether this tool has the same CLI arguments as gofmt?

Formatting has to be "pick only one per language", and right now we say you can drop in anything compatible (https://github.com/aspect-build/rules_lint/blob/main/docs/format.md#multi_formatter_binary-go)
so as long as that's the case, I think this is close to being the right shape.

example/tools/BUILD
```diff
@@ -132,6 +141,7 @@ shellcheck_binary(name = "shellcheck")
 multi_formatter_binary(
     name = "format",
     cc = "@llvm_toolchain_llvm//:bin/clang-format",
+    # Or ":gofumpt" if you want to use gofumpt instead of gofmt
```
alexeagle, line 144: it's a bit rough that we have no way to test gofumpt is working at all. Maybe we need another `multi_formatter_binary` in the test folder.

alexeagle (approved):

example/tools/BUILD
jelmer: I can add something like this:

```
multi_formatter_binary(
    name = "format_fumpt",
    go = ":gofumpt"
)
```

is that actually going to be exercised by anything though?

alexeagle: No, and I don't see any test fixture we could extend under the example/ folder either. This is fine already, it's so simple there's not much surface area where it could have a bug.

PR #118, opened 2024-01-25
fix early glob expansion, starlark patterns

### LINKED ISSUE #117: [Bug]: Starlark formatter no longer touches BUILD files
### What happened?

```console
BUILD.bazel
737
```

### Version

### How to reproduce

```shell
```

### Any other information?

```
...
++ case "$language" in
++ '[' 0 -eq 0 ']'
...
```

pjjw:
tl;dr this is a partial correction of #117 

Looks like the patterns in the reworked format.sh are not catching everything anymore, partially because of an unquoted bash array expansion, and partially because the patterns are incomplete as written here.

This is an incomplete fix, so don't expect this to be merged exactly (don't have more time today), but at least wanted to get this reported and put up this fix for at least the part i'm concerned about

More work is probably needed for other languages, as the patterns look to be probably deficient in the same way the starlark patterns are
---

### Type of change

- Bug fix

### DIFF
--- a/format/private/format.sh
+++ b/format/private/format.sh
@@ -62,7 +62,7 @@ function ls-files {
 
         # TODO: determine which staged changes we should format; avoid formatting unstaged changes
         # TODO: try to format only modified regions of the file (where supported)
-        git ls-files --cached --modified --other --exclude-standard ${patterns[@]} | {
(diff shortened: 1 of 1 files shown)


pjjw:
obsoleted by #120

PR #119, opened 2024-01-26
collapse common logic into run-format formula

pjjw:
took a crack at parallelizing the formatter script, i think this works OK.

---

### Type of change

- Performance (a code change that improves performance)

### Test plan
- Patched this into another repository where rules_lint is in use, kicked the tires, made sure output was the same, it ran faster, and deviations were caught.

### DIFF
--- a/format/private/format.sh
+++ b/format/private/format.sh
@@ -14,10 +14,16 @@ cd $BUILD_WORKSPACE_DIRECTORY
 
 function on_exit {
   code=$?
-  if [[ $code != 0 ]]; then
-    echo >&2 "FAILED: A formatter tool exited with code $code"
-    echo >&2 "Try running 'bazel run {{fix_target}}' to fix this."
-  fi
+  case "$code" in
+    # return code 143 is the result of SIGTERM, which isn't failure, so suppress failure suggestion
+    0|143)
+      exit $code;
+      ;;
+    *)
+      echo >&2 "FAILED: A formatter tool exited with code $code"
+      echo >&2 "Try running 'bazel run {{fix_target}}' to fix this."
+      ;;
+  esac
 }
 
 trap on_exit EXIT
@@ -61,7 +67,7 @@ function ls-files {
         exit 1
         ;;
     esac
-    
+
     if [ "$#" -eq 0 ]; then
         # When the formatter is run with no arguments, we run over "all files in the repo".
         # However, we want to ignore anything that is in .gitignore, is marked for delete, etc.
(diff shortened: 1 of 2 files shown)


IamXander:
I would not call myself a bash expert by any means but if what you said in the title works this is an amazing change and I would love to see it merged.

format/private/format.sh
```diff
@@ -45,7 +51,7 @@ function ls-files {
       'SQL') patterns=('*.sql' '*.cql' '*.ddl' '*.inc' '*.mysql' '*.prc' '*.tab' '*.udf' '*.viw') ;;
       'Scala') patterns=('*.scala' '*.kojo' '*.sbt' '*.sc') ;;
       'Shell') patterns=('.bash_aliases' '.bash_functions' '.bash_history' '.bash_logout' '.bash_profile' '.bashrc' '.cshrc' '.flaskenv' '.kshrc' '.login' '.profile' '.zlogin' '.zlogout' '.zprofile' '.zshenv' '.zshrc' '9fs' 'PKGBUILD' 'bash_aliases' 'bash_logout' 'bash_profile' 'bashrc' 'cshrc' 'gradlew' 'kshrc' 'login' 'man' 'profile' 'zlogin' 'zlogout' 'zprofile' 'zshenv' 'zshrc' '*.sh' '*.bash' '*.bats' '*.cgi' '*.command' '*.fcgi' '*.ksh' '*.sh.in' '*.tmux' '*.tool' '*.trigger' '*.zsh' '*.zsh-theme') ;;
-      'Starlark') patterns=('BUCK' 'BUILD' 'BUILD.bazel' 'MODULE.bazel' 'Tiltfile' 'WORKSPACE' 'WORKSPACE.bazel' '*.bzl' '*.star') ;;
+      'Starlark') patterns=('BUCK' '**/BUCK' 'BUILD' '**/BUILD' 'BUILD.bazel' '**/BUILD.bazel' 'MODULE.bazel' 'Tiltfile' '**/Tiltfile' 'WORKSPACE' 'WORKSPACE.bazel' '*.bzl' '*.star') ;;
```
IamXander, line 54: You will be able to get rid of the "**" and replace it with "*" once https://github.com/aspect-build/rules_lint/pull/124 lands for consistency.

pjjw: the `**` isn't to prevent the accidental glob expansion by bash addressed in #124 (that problem is fixed on line 71 by wrapping the pattern array expansion in quotes), but to get `git ls-files` to recurse in searches. there does appear to be some inconsistencies in pattern matching in git- this is what i found to work correctly.

pjjw:
updated to pass tests

pjjw:
@alexeagle got a moment to give this a once-over, or at least let me know if there's interest?

alexeagle:
Yes! sorry for the radio silence, this is indeed an important improvement.

How is the output handled? My biggest concern is that we shouldn't interleave outputs in a confusing way, rather each background job has to buffer and then we print the buffer after that job has completed.

format/private/format.sh
```diff
@@ -14,10 +14,15 @@ cd $BUILD_WORKSPACE_DIRECTORY
 
 function on_exit {
   code=$?
-  if [[ $code != 0 ]]; then
-    echo >&2 "FAILED: A formatter tool exited with code $code"
-    echo >&2 "Try running 'bazel run {{fix_target}}' to fix this."
-  fi
+  case "$code" in
+    0|143)
```
alexeagle, line 18: what is code 143? needs a comment

format/private/format.sh
```diff
@@ -123,160 +128,79 @@ case "$mode" in
...
+        else
+          ( echo "$files" | tr \\n \\0 | xargs -0 "$bin" $args >&2 ; times ) | ( read _ _; read tuser tsys; echo "Formatted ${lang} in ${tuser}" )
+        fi
+        ;;
+      Java)
+          # Setting JAVA_RUNFILES to work around https://github.com/bazelbuild/bazel/issues/12348
+          ( echo "$files" | tr \\n \\0 | JAVA_RUNFILES="${RUNFILES_MANIFEST_FILE%_manifest}" xargs -0 "$bin" $args >&2 ; times ) | (read _ _ ;   read tuser tsys; echo "Formatted ${lang} in ${tuser}" )
+        ;;
+      Swift)
+        # for any linter that must be silenced
```
alexeagle, line 170: nit: linter -> formatter ?

format/private/format.sh
```diff
@@ -123,160 +128,79 @@ case "$mode" in
...
+            echo "Go files not formatted:"
+            echo "$NEED_FMT"
+            exit 1
+          fi
+        else
+          ( echo "$files" | tr \\n \\0 | xargs -0 "$bin" $args >&2 ; times ) | ( read _ _; read tuser tsys; echo "Formatted ${lang} in ${tuser}" )
+        fi
+        ;;
+      Java)
+          # Setting JAVA_RUNFILES to work around https://github.com/bazelbuild/bazel/issues/12348
```
alexeagle, line 166: nit: I wonder if it's possible to reduce the duplication, either by setting `JAVA_RUNFILES` in the default case, or setting it and then falling through to the default case

format/private/format.sh
```diff
@@ -123,160 +128,79 @@ case "$mode" in
...
-files=$(ls-files 'Protocol Buffer' $@)
-bin=$(rlocation {{buf}})
-if [ -n "$files" ] && [ -n "$bin" ]; then
-  echo "Formatting Protobuf with buf..."
-  for file in $files; do
-    $bin $bufmode $file
-  done
-fi
+(
+  trap "kill 0" SIGINT
```
alexeagle, line 183: could we break this up into two PRs please?

Let's land the refactoring to introduce the `run-lints` function first, but not run them in parallel yet. That way you don't get stuck with any messy rebases while we figure out how to buffer the output properly.

alexeagle (commented):
pulled this down and tried it, and yeah the output is interleaved in a way that I don't think is desirable.

pjjw:
got any suggestions as to how you'd like it to look?

alexeagle:
I think we can't print anything like "Formatting x with y..." because those will all appear at the top as the subshells are spawned in the background.

And when a tool completes I think we want to print the full buffer it produced.
I'll look around a bit for some example we can follow.

alexeagle:
Right, so the simple thing in Bash is to redirect the stdout/stderr of the individual tools to a file. Then when the tool is finished we can `cat` the file along with the line we print about what we ran and how long it took.

It's tempting to rewrite it in a language where there's a nice terminal experience available in some library, like superconsole. It would be a big effort though.

alexeagle:
sorry, merge conflicts now. would be nice to get the refactor in first :)

pjjw:
busy week, sorry, will look later today

format/private/format.sh
pjjw: i pulled some of the timing logic into another helper function, where possible. PTAL

alexeagle: that's plenty of improvement, thanks!

alexeagle (approved):

PR #120, opened 2024-01-26
fix filename pattern matching

### LINKED ISSUE #117: [Bug]: Starlark formatter no longer touches BUILD files

pjjw:
git ls-files seems to match against filename strings exactly, and doesn't use the same logic as e.g. gitignore. this means that to match files by exact filename below the root, we need a */ at the start of the pattern to match more than files strictly at the root of the tree. these prefixed patterns however miss the root, so we need to include both forms.

additionally, the pattern array needs to be quoted to prevent the shell from doing glob expansion against any files in the current directory.

Fixes #117 and #122.

---

### Type of change

### DIFF
--- a/format/private/format.sh
+++ b/format/private/format.sh
@@ -34,18 +34,18 @@ function ls-files {
       'Go') patterns=('*.go') ;;
       'HCL') patterns=('*.hcl' '*.nomad' '*.tf' '*.tfvars' '*.workflow') ;;
       'HTML') patterns=('*.html' '*.hta' '*.htm' '*.html.hl' '*.inc' '*.xht' '*.xhtml') ;;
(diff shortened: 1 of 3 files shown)


IamXander:
I totally missed this before I submitted https://github.com/aspect-build/rules_lint/issues/122 and https://github.com/aspect-build/rules_lint/pull/124. Both of these fix the same problem but in https://github.com/aspect-build/rules_lint/pull/124 might be a little nicer since you don't have to duplicate globs to include a "**" version.

alexeagle:
Thanks both of you, this is my fault for not having more tests in place for the formatter before #105 . I'm going to work on adding those first so we don't keep chasing regressions.

pjjw:
@IamXander the problem isn't fully fixed by #124, though- failing to recurse for e.g. `BUILD` files isn't caused by inadvertent glob expansion, but the (kinda inconsistent) semantics of `git ls-files`: if there's no wildcard, it's exact-match only.

format/private/format.sh
```diff
@@ -54,15 +54,15 @@ function ls-files {
...
+
     if [ "$#" -eq 0 ]; then
         # When the formatter is run with no arguments, we run over "all files in the repo".
         # However, we want to ignore anything that is in .gitignore, is marked for delete, etc.
         # So we use `git ls-files` with some additional care.
 
         # TODO: determine which staged changes we should format; avoid formatting unstaged changes
         # TODO: try to format only modified regions of the file (where supported)
-        git ls-files --cached --modified --other --exclude-standard ${patterns[@]} | {
+        git ls-files --cached --modified --other --exclude-standard "${patterns[@]}" | {
```
pjjw, line 65: @IamXander adding in a ` | tee /dev/stderr ` here is helpful for debugging, doing the same in #124 will show that even there only the root BUILD file is being considered without the add'l pattern expansion

alexeagle:
Replaced by #127

Issue #121, opened 2024-01-27
[Bug]: JSON format does not seem to be supported


IamXander:
### What happened?

There is no way to format json files using this tool. It looks like there is scaffolding for it in `format.sh` but then it is never called.

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: bazel 6.3.2

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: rules_lint-0.10.0

Language(s) and/or frameworks involved: json

### Any other information?

I will hopefully get to submitting a PR myself but in case I do not I wanted to put this issues out there in case another contributor beat me to it.

PR #123, opened 2024-01-27
fix(formatter): json needs a distinct tool entry

IamXander:
This fixes https://github.com/aspect-build/rules_lint/issues/121 in a similar way to how this sort of thing was fixed in https://github.com/aspect-build/rules_lint/pull/99
---

### Type of change

### Test plan

Verified that with JSON is now being formatting using my fork and branch!

### DIFF
--- a/docs/format.md
+++ b/docs/format.md
@@ -37,8 +37,8 @@ multi_formatter_binary(
 ## multi_formatter_binary
 
 <pre>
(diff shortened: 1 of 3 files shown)


alexeagle:
Thanks, it's extra embarrassing since I had that bug in mind when I did the refactor to make json its own language, but convinced myself it was working. Obviously this needs a new test fixture that checks formatting works one language at a time.

IamXander:
Any objection to merging this while you work on the test harness framework? I'll add a test based on how you want to structure the unit testing.

IamXander:
Ah I see I made a mistake! Ran `format` so it should pass now

alexeagle:
Making it a separate tool is one way to resolve this, but the simpler way is to copy from CSS and HTML handling. I added coverage for it to #126 which I think resolves #121 in place of this PR, wdyt?

IamXander:
Happy to close this in favor of https://github.com/aspect-build/rules_lint/pull/126. Ngl makes me a little sad that my name is not making it into the repo ;). However, its totally fine. Just happy to see it fixed.

alexeagle:
I have no doubt we'll have more of your commits in the history! :)

Issue #122, opened 2024-01-27
[Bug]: The bash escaping in format.sh will stop recursing if matched at root level


IamXander:
### What happened?

I am using the markdown formatter and I have many markdown files in my repo. However, when I run `bazel run //:format` it will only find the markdown files in the top level.

```
# This formats only README_TOP_LEVEL.md
WORKSPACE
README_TOP_LEVEL.md
subfolder1/
  README_NOT_FORMATTED1.md
subfolder2/
  README_NOT_FORMATTED2.md
```

Interestingly, when I delete `README_TOP_LEVEL.md` (shown below) the formatter works.
```
# This formats all files
WORKSPACE
subfolder1/
  README_NOT_FORMATTED1.md
subfolder2/
  README_NOT_FORMATTED2.md
```

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: bazel 6.3.2

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: git version 2.41.0

Language(s) and/or frameworks involved:
git version 2.41.0

### How to reproduce

Run the formatter with a repo that looks like this
```shell
# This formats only README_TOP_LEVEL.md
WORKSPACE
README_TOP_LEVEL.md
subfolder1/
  README_NOT_FORMATTED1.md
subfolder2/
  README_NOT_FORMATTED2.md
```

### Any other information?

I was able to narrow it down to some bash expansions. Here is a minimal reproducer. 
```shell
patterns=('contents.lr' '*.md' '*.livemd' '*.markdown')
git ls-files --cached --modified --other --exclude-standard ${patterns[@]} | {
  grep -vE "^$(git ls-files --deleted)$" || true;
}
```
By adding some echos I think I narrowed it down to some bash magic where the single quotes are escaped.
```shell
patterns=('contents.lr' '*.md' '*.livemd' '*.markdown')
echo "git ls-files --cached --modified --other --exclude-standard ${patterns[@]}"
#Prints
#git ls-files --cached --modified --other --exclude-standard contents.lr *.md *.livemd *.markdown
git ls-files --cached --modified --other --exclude-standard ${patterns[@]} | {
  grep -vE "^$(git ls-files --deleted)$" || true;
}
```
I assume that should have some single quotes separating '*.md' '*.livemd' '*.markdown'

Finally I tried
```shell
patterns=('contents.lr' '*.md' '*.livemd' '*.markdown')
[...]

PR #124, opened 2024-01-27
Disabled globbing in format.sh

IamXander:
By disabling bash globbing patterns like *.md are passed directly to git commands. Without disabling globbing *.md would be expanded to README.md before being passed to git. This fixes an issue where having a README.md in the top level would prevent recursive formatting.

See issue https://github.com/aspect-build/rules_lint/issues/122

---

### Type of change

### Test plan

Tested in a repo that mirrors the repo described in https://github.com/aspect-build/rules_lint/issues/122 and was able to recursively find *.md files.

### DIFF
--- a/format/private/format.sh
+++ b/format/private/format.sh
@@ -22,6 +22,10 @@ function on_exit {
 
 trap on_exit EXIT
 
+# Disable globbing, disabling globbing will allow formatting even if there is a match at the top level
+# See https://github.com/aspect-build/rules_lint/issues/122
+set -o noglob
+
 # ls-files <language> [<file>...]
 function ls-files {
     language="$1" && shift;


alexeagle:
I've got a draft in #126 to add the missing test coverage, and I patched this simple fix there. It does not seem to be sufficient on its own: https://github.com/aspect-build/rules_lint/actions/runs/7681777431/job/20935308636?pr=126

I'll keep debugging it for a bit.

IamXander:
Just looked at your changes in #126 and the test case seems right to me so I agree there is something more going on with this this. I'd be happy to dig in after you merge your test case.

Not to make things too complex but have you considered doing the format.sh script in python (or a non bash scripting language)?

alexeagle:
I did consider another language. Python has the downside of needing a runtime which not everyone has setup. (it was a big problem with rules_pkg among others)
https://github.com/aspect-build/rules_lint/pull/105 deleted the TODO about "maybe it should be rewritten" and at this point I think we're committed.

Shell (and Bash as an extension) is actually a pretty great language, especially when most of the work is spawning subprocesses and plumbing stdio and argv. Not many people know it deeply though!

alexeagle:
replaced by #127

PR #126, opened 2024-01-27
ci: add test coverage for format binary

### LINKED ISSUE #121: [Bug]: JSON format does not seem to be supported

alexeagle:
I suggested this in https://github.com/aspect-build/rules_lint/commit/18ca19086309ef587ab22dde09db3135cb79f7d0

> Note, there ought to be more testing that would have caught this. I spent some time writing a bats test here 

Should be able to catch most regressions this way.

Fixes #121 

---

### Type of change

### Test plan

- New test cases added


format/private/format.sh
```diff
@@ -140,6 +140,13 @@ if [ -n "$files" ] && [ -n "$bin" ]; then
   echo "$files" | tr \\n \\0 | xargs -0 $bin $prettiermode
 fi
 
+files=$(ls-files JSON $@)
```
alexeagle, line 143: @IamXander I think this is the fix for #121

IamXander: This looks right to me. Originally it was nice to have this as a separate tool so I could introduce this to my company's repo language by language. I would just turn it on for JSON, then I would enable it on more things. However, that is not a big deal so I am totally fine with this.

alexeagle: I agree that still might make sense in a follow-up PR, especially if there's an alternative tool for JSON (I removed YAML from the prettier scope a while ago because someone said it's not the best choice there)

In fact, we should maybe have a 1:1 mapping between the language attributes on `multi_formatter_binary` and the list of languages in the GitHub Linguist sync (https://github.com/aspect-build/rules_lint/blob/main/format/private/filter.jq#L3-L22) - and that should happen before this repo can go to 1.0.

Would love to have more contributions :)

IamXander: After this merges should I reopen https://github.com/aspect-build/rules_lint/pull/123/files to give json and other a distinct tool entry? Is that your suggestion?

gregmagolan (approved):
🌯

format/private/format.sh
alexeagle: I'm not sure. The downside is that a user will have repeat `lang1: prettier, lang2: prettier, etc..` attributes which doesn't seem very ergonomic either. Let's start with an issue for it so there's a place to discuss.

PR #127, opened 2024-01-28
fix filename pattern matching

### LINKED ISSUE #117: [Bug]: Starlark formatter no longer touches BUILD files

pjjw:
found a less-intrusive way to do this vs #120

Fixes #117 and #122.

---

### Type of change

### Test plan

- Passes tests in preliminary #126

### DIFF
--- a/format/private/format.sh
+++ b/format/private/format.sh
@@ -62,7 +62,7 @@ function ls-files {
 
         # TODO: determine which staged changes we should format; avoid formatting unstaged changes
         # TODO: try to format only modified regions of the file (where supported)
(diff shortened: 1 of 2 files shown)


alexeagle (approved):
Yeah there we go, very precise :)
Thanks!

I rebased to pick up and update the new tests which caught more missing files than I realized.

PR #133, opened 2024-01-31
format: fix exit code of check when using clang-format

tomgr:
### Type of change

- Bug fix

- Breaking change (this change will force users to change their own code or config): No
- Relevant documentation has been updated: no relevant documentation

fixed the exit code of `multi_formatter_binary(...)` when `--mode check` is used with clang-format.

### Test plan

Manual testing:

- Modified example/hello.cpp to introduce a formatting error (e.g. remove a space after the left brace).
- Ran `bazel run //tools:format -- --mode check` and check the exit code is 1.
- Fixed the formatting error via `bazel run //tools:format`
- Ran `bazel run //tools:format -- --mode check` and checked the exit code is 0.

### DIFF
--- a/format/private/format.sh
+++ b/format/private/format.sh
@@ -101,7 +101,7 @@ case "$mode" in
    tfmode="-check -diff"
    jsonnetmode="--test"
    scalamode="--test"
-   clangformatmode="--style=file --fallback-style=none --dry-run"
+   clangformatmode="--style=file --fallback-style=none --dry-run -Werror"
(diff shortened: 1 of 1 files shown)


alexeagle (approved):
Thanks!

If you're interested, we should add negative test cases in https://github.com/aspect-build/rules_lint/blob/main/format/test/format_test.bats#L116
they are all assert_success but we should have `assert_failure` cases in there too.

PR #136, opened 2024-02-05
chore: remove Go

alexeagle:
We don't have any Go sources in rules_lint so there's no reason we should install a Go toolchain or worry about upgrading it. 'aspect configure' provides the auto-generation for bzl_library targets so we don't need Gazelle either.


alexeagle:
This repo has no aspect workflows setup, so there's no CI coverage for `bzl_library` targets being updated.
Will add a follow-up task.

PR #139, opened 2024-02-07
chore: cleanup maybe patterns

gregmagolan:
Just some minor cleanup on the maybe patterns. In most places we don't have `name` so this makes the patterns here consistent with our other repos.

### DIFF
--- a/lint/private/maybe.bzl
+++ b/lint/private/maybe.bzl
@@ -1,0 +1,7 @@
+"maybe utilities"
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", _http_archive = "http_archive")
+load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
+
+def maybe_http_archive(**kwargs):
+    maybe(_http_archive, **kwargs)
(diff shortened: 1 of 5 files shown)


alexeagle (approved):

alexeagle:
Can't wait for Marvin to suggest the docs fix that made the PR red :)

alexeagle (approved):

lint/private/maybe.bzl
```diff
@@ -0,0 +1,7 @@
+"maybe utilities"
```
alexeagle, line 1: I really don't like this pattern, what other "maybe utilities" are there? I think it was better when this two-line function was inlined at the use site

gregmagolan: It ends up being 4 lines since there are two load statements vs. one line to just load the maybe. My preference is the one line since it seems wasteful to copy the same 4 lines everywhere you need the maybe.

To answer your question of "what other maybe utilities", some repos need `maybe_http_file` as well and I've seen a few other variants. A `private/maybe.bzl` is a home for that boilerplate so use sites still need just one line if there are mutiple variants vs 4 lines X <number of variants>.

gregmagolan: bzlmod will make all this go away in the near future one way or the other

gregmagolan: unrelated to maybe, can you tell why this PR CI is red for the one e2e test???

jbedard (approved):

PR #144, opened 2024-02-09
docs(golangci-lint): fix the target name

srabraham:
This is what the //example workspace uses, and it's what I had to use to get this to work in our VideoAmp repo.
https://github.com/aspect-build/rules_lint/blob/ebe8678587abb1f9104abbfc6ca2ee1a8579b4aa/example/tools/lint.bzl#L59-L62

---

### Type of change

- Documentation (updates to documentation or READMEs)

### Test plan

- This documentation matches the working version in //example


alexeagle (approved):
Thanks!!

PR #159, opened 2024-02-22
refactor: break up tools/ into tools/format and tools/lint

alexeagle:
Suggested during API review, as this encourages users to have more folders underneath tools/ to avoid eager fetching more than needed for a given target. For our own example, it doesn't really matter. But we want new users to go down a better path.

---

### Type of change

### Test plan


example/tools/format/BUILD
```diff
@@ -4,15 +4,10 @@ This is in its own package because it has so many loading-time symbols,
 we don't want to trigger eager fetches of these for builds that don't want to run format.
 """
 
-load("@aspect_bazel_lib//lib:copy_to_directory.bzl", "copy_to_directory")
 load("@aspect_rules_lint//format:defs.bzl", "multi_formatter_binary")
```
gregmagolan, line 7: nit: this file is `example/tools/format/BUILD.bazel` in the docs above

example/tools/lint/BUILD
```diff
@@ -0,0 +1,87 @@
+"""BUILD definition for the formatter binary
```
gregmagolan, line 1: nit: BUILD.bazel for consistency?

gregmagolan (approved):
🌮

PR #160, opened 2024-02-23
Use multirun

alexeagle:
This has the side-effect of making a formatter binary per-language, which was one of our design review outcomes, since Aspect Workflows might want to be able to run them individually or in parallel.

FYI @keith since it's taking a dependency on your repo

Also now formats-in-parallel using the `jobs` attribute of `multirun`.

### Type of change

`multi_formatter_binary` is renamed to `format_multirun` and some attributes were renamed: 

- `sh` is now `shell`
- `protobuf` is now `protocol_buffer`.

### Test plan

### DIFF
--- a/format/private/format.sh
+++ b/format/private/format.sh
@@ -1,9 +1,17 @@
 #!/usr/bin/env bash
-# Template file for a formatter binary. This is expanded by a Bazel action in formatter_binary.bzl
-# to produce an actual Bash script.
-# Expansions are written in "mustache" syntax like {{interpolate_this}}.
+# Wrapper around a formatter tool
 
-{{BASH_RLOCATION_FUNCTION}}
+# --- begin runfiles.bash initialization v3 ---
+# Copy-pasted from the Bazel Bash runfiles library v3.
+# https://github.com/bazelbuild/bazel/blob/master/tools/bash/runfiles/runfiles.bash
+set -uo pipefail; set +e; f=bazel_tools/tools/bash/runfiles/runfiles.bash
+source "${RUNFILES_DIR:-/dev/null}/$f" 2>/dev/null || \
+  source "$(grep -sm1 "^$f " "${RUNFILES_MANIFEST_FILE:-/dev/null}" | cut -f2- -d' ')" 2>/dev/null || \
+  source "$0.runfiles/$f" 2>/dev/null || \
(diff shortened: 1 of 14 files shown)


format/private/format.sh
```diff
@@ -154,15 +117,13 @@ function time-run {
...
   local lang="$1" && shift
-  local fmtname="$1" && shift
   local bin="$1" && shift
   local args="$1" && shift
   local tuser
   local tsys
 
   local files=$(ls-files "$lang" $@)
   if [ -n "$files" ] && [ -n "$bin" ]; then
-    echo "Formatting ${lang} with ${fmtname}..."
```
jbedard, line 165: Was this replaced with something?

alexeagle: Yes, multi run prints the label of the target it's about to run

alexeagle: Advantage is that when we buffer stdout of this tool we don't lose the print statement at the start time

format/test/format_test.bats
```diff
@@ -6,151 +6,130 @@ bats_load_library "bats-assert"
 
 # No arguments: will use git ls-files
 @test "should run prettier on javascript using git ls-files" {
-    run bazel run //format/test:format_javascript
+    run bazel run //format/test:format_JavaScript_with_prettier
     assert_success
 
-    assert_output --partial "Formatting JavaScript with Prettier..."
```
jbedard, line 12: and this wasn't replaced with something equivalent?

alexeagle: No, because this test doesn't execute the `multirun` which prints `Running @@//tools/format:format_Markdown_with_prettier`. Instead it calls that one binary directly.

alexeagle: Oh and look, Keith is helping us out :)
https://github.com/keith/rules_multirun/pull/35

alexeagle:
This introduces a dep on rules_python, purely to run a `py_binary`. We kinda hate that so adding @thesayyn to make any alternative suggestions :)

thesayyn:
my take on this is that since we don't allow users to pass args per binary like multirun, we can get away with a simple bash script it seems...  i don't like rules_python dependency here, eventhough it's indirect.

alexeagle:
Simple bash script cannot run the tools in parallel without interleaving their stdouts though. So we need something more.

thesayyn (approved):

PR #161, opened 2024-02-24
feat: use rules_multitool

alexeagle:
Suggestions from 1.0 API design review.
I'll followup with other provided tools in a subsequent PR after we are happy with the shape of this one.

- bzlmod users no longer have a fetch that's still hanging out in WORKSPACE.bzlmod
- the name of the repo is prefixed with aspect_rules_lint. so we don't worry about collisions

---

### Type of change

### Test plan


multitool/defs.bzl
```diff
@@ -0,0 +1,54 @@
+"""fork of rules_multitool
```
jbedard, line 1: Is this public API or just an internal util? Can we break it later being the real question...

multitool/defs.bzl
```diff
@@ -0,0 +1,54 @@
...
+Derived from https://github.com/theoremlp/rules_multitool
+but simplified to just generate an alias without any toolchains.
+"""
+
+_COMMON_FILES_TO_GENERATE = [
+    "BUILD.bazel",
+    "WORKSPACE.bazel",
+]
+
+_PLATFORMS = [
```
jbedard, line 11: This implies all tools have these 4 platforms and only these? ~~If~~ When we have a tool that has more/less will it require a change to a public API? (if `multitool/defs.bzl` is non-public then this doesn't matter atm)

alexeagle: bazel-lib already has a bunch of "download a tool and make it a toolchain" so maybe there's a good reason to make a side effort here to consolidate the options and have a way that scales to "lots" of tools.

alexeagle:
FYI @mark-thm

multitool/defs.bzl
alexeagle: trying again with rules_multitool. It still has only these four platforms, but I think we can add more without a breaking change. Just means a platform like windows was previously an error, and then can start working.

mark-thm: Happy to take a PR to extend os/architecture set -- recent changes should make the architectures pretty straight forward. We don't really have a setup to test on Windows, so might be a bigger lift to add.

alexeagle: We can contribute for Windows, thanks!

alexeagle:
Oh there's still a bug in rules_multitool. I see that I'm downloading tools for *all* platforms rather than only for the host platform. @mark-thm want me to file an issue on the repo?

mark-thm:
@alexeagle it's unclear to me what the right behavior is here, but I threw together https://github.com/theoremlp/rules_multitool/pull/12 as a candidate fix.

alexeagle:
That bug was fixed in https://github.com/theoremlp/rules_multitool/pull/16 (thanks!) so this should be ready to go @jbedard

jbedard:
Have you looked at the CI failures?

mark-thm:
You need 0.4.0 to resolve those (it contains the fix that allows you to have a top-level `$schema` key)

jbedard (approved):

PR #164, opened 2024-03-08
refactor: use an explicit flag rather than attr_attribute

alexeagle:
The latter is a global namespace, and also leaked into a test where it shouldn't have been needed

---

### Type of change

`--aspects_parameters=fail_on_violation=true` must be replaced with `--@aspect_rules_lint//lint:fail_on_violation`

### Test plan

Ran `lint.sh --fail-on-violation //src:all` in the example.


thesayyn (approved):
Nice! looks much better

PR #165, opened 2024-03-08
chore: upgrade rules_proto

alexeagle:
Also replace com_google_protobuf with toolchains_protoc so we aren't causing ppl to compile `protoc` :)

### Type of change

### Test plan


thesayyn (approved):
Awesome

PR #183, opened 2024-03-27
chore: upgrade rules_multitool for WORKSPACE users

alexeagle:
Includes https://github.com/theoremlp/rules_multitool/pull/21 which we want in aspect silo

Note, it's not necessary to increase the lower bound for bzlmod users, as that API already allowed multiple lockfiles.

---

### Type of change

Now requires bazel_features as a transitive WORKSPACE dep of rules_multitool.

### Test plan


thesayyn (approved):

alexeagle:
that's strange, why did it merge on red...

PR #184, opened 2024-03-27
chore: fix ci, need newer bazel_features version

alexeagle:


format/repositories.bzl
```diff
@@ -29,6 +29,14 @@ def rules_lint_dependencies():
         url = "https://github.com/theoremlp/rules_multitool/releases/download/v0.6.0/rules_multitool-0.6.0.tar.gz",
     )
 
+    # Transitive of rules_multitool, included here for convenience
```
jbedard, line 32: Shouldn't we be calling `multitool_deps()` somewhere to do this? Or do we have no `rules_lint_deps()` to put it in?

jbedard (approved):

PR #187, opened 2024-03-28
chore: add terraform_linux_aarch64

hunshcn:
---

### Type of change

### Test plan


alexeagle (approved):
thanks!

PR #188, opened 2024-03-28
feat: provide a default for Terraform formatter

alexeagle:
Also give users control of the 'jobs' parameter

---

### Type of change

BREAKING CHANGES:
- fetch_terraform() has been removed, get it from @multitool instead

### Test plan


thesayyn (approved):
🍓

Issue #175, opened 2024-03-21
[FR]: make formatter's --mode check to be a test rule


hunshcn:
### What is the current behavior?

```
bazel run //:format -- --mode check
```
we can only use `bazel run` to run check. why not make it to be a test.

Cooperating with https://github.com/aspect-build/rules_lint/issues/97 will be very helpful in ci.

### Describe the feature

```
bazel test //:format_test
```
```bazel
format_multirun_test(
    name = "format_test",
    go = "@cc_mvdan_gofumpt//:gofumpt",
    starlark = "@com_github_bazelbuild_buildtools//buildifier",
)
```

alexeagle:
It cannot be a test because that would make a node in the dependency graph that "depends on everything in the repo".

In fact, the formatter binary doesn't even require that files to be formatted are listed as `srcs` of any Bazel target.
I suggest that CI run the formatter outside of Bazel. Note Aspect Workflows has first-class support: https://docs.aspect.build/workflows/configuration/lint

hunshcn:
srcs is not required, wo can use a workspace arg (//:MODULE.bazel) to resolve WORKSPACE path

hunshcn:
that's https://github.com/keith/buildifier-prebuilt do.

hunshcn:
I think it would be cool to have all the tests in bazel test //...
And it's really possible to implement this, all ready-made, isn't it?
If you agree with me, I can implement this feature.

alexeagle:
Bazel tests are meant to declare their inputs, so they are hermetic and can therefore be correctly cached.

hunshcn:
Can we refer to https://github.com/bazelbuild/buildtools/pull/1092? only accept no-cache.

hunshcn:
Working with [kazel](https://github.com/kubernetes/repo-infra/tree/master/cmd/kazel) is also a way. Then set `srcs = "//:all-files"

hunshcn:
> Can we refer to [bazelbuild/buildtools#1092](https://github.com/bazelbuild/buildtools/pull/1092)? only accept no-cache.



> Working with [kazel](https://github.com/kubernetes/repo-infra/tree/master/cmd/kazel) is also a way. Then set `srcs = "//:all-files"`
> 
> Of course, this also has a disadvantage (you need to add an additional kazel test (no-cache)).
> 
[...]

PR #189, opened 2024-03-29
feat: format_test

hunshcn:
Resolve https://github.com/aspect-build/rules_lint/issues/175

---

### Type of change

- Documentation (updates to documentation or READMEs)

```
new rule `format_test` is a available to create format test. see docs/format.md for more infomation.
```

### Test plan

- New test cases added
Because the wrong format file cannot be imported into the repo, the failure test of `workspace mode` can only be performed manually. (Manually change some code to the wrong format.)

### DIFF
--- a/docs/formatting.md
+++ b/docs/formatting.md
@@ -88,6 +88,26 @@ $ chmod u+x .git/hooks/pre-commit
 
 ### Check that files are already formatted
 
+There are two ways.
+
+#### one: manual check
+
 This will exit non-zero if formatting is needed. You would typically run the check mode on CI.
 
 `bazel run //tools/format:format.check`
+
+#### two: test rule
+
+```starlark
+load("@aspect_rules_lint//format:defs.bzl", "format_test")
+
+format_test(
+    name = "format_test",
+    # register languages, e.g. same to format_multirun
+    # python = "//:ruff",
+    workspace = "//:WORKSPACE.bazel", # not cacheable
+    # srcs = ["files"], # cacheable, but need declare all files. recommend use with kazel
+)
+```
+
+Then run `bazel test //tools/format/...` to check that all files are formatted.
--- a/docs/formatting.md
+++ b/docs/formatting.md
@@ -90,24 +90,49 @@ $ chmod u+x .git/hooks/pre-commit
 
 There are two ways.
 
-#### one: manual check
+#### 1: `run` target
 
 This will exit non-zero if formatting is needed. You would typically run the check mode on CI.
 
 `bazel run //tools/format:format.check`
 
-#### two: test rule
+#### 2: `test` target
+
+Normally Bazel tests should be hermetic, declaring their inputs, and therefore have cacheable results.
+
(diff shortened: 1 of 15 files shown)


hunshcn:
@alexeagle

alexeagle:
Ugh, the test failure looks like something latent that this has tickled :(

hunshcn:
C++_with_clang-format passed locally, let me open a codespace.

`Java_with_java-format` and `Kotlin_with_ktfmt` failed even at main branch in my m1 mac and arm64 linux. I have no idea why

hunshcn:
in codespace, all passed.........(The swiftformat failed by so file is missing, which is reasonable.)

image

hunshcn:
clang need libtinfo.so.5, so we need update ci and install this dep before test.

bazel-contrib/.github/.github/workflows/bazel.yaml@v5 does not support. v6 is required ( configurate bazel_test_command), but new problems happen.

Java is still broken, I don't know why, and I don't understand it.

hunshcn:
if turn off clang, revert changes of ci.yaml

docs/formatting.md
```diff
@@ -88,6 +88,26 @@ $ chmod u+x .git/hooks/pre-commit
...
+
+```starlark
+load("@aspect_rules_lint//format:defs.bzl", "format_test")
+
+format_test(
+    name = "format_test",
+    # register languages, e.g. same to format_multirun
+    # python = "//:ruff",
+    workspace = "//:WORKSPACE.bazel", # not cacheable
+    # srcs = ["files"], # cacheable, but need declare all files. recommend use with kazel
```
alexeagle, line 109: note, I've never heard of `kazel` before, most Bazel users are familiar with Gazelle for build file generation. I'll update this.

hunshcn: ok

alexeagle:
Hm, I think we should probably have the same API as `buildifier_test` with https://github.com/bazelbuild/buildtools/pull/1092/files#diff-c5e2ba5db338e88cf7c850afd25c044dfa97198fefa4b296370c9286ab81d415R97 in addition to `srcs`/`workspace`, any thoughts about that @hunshcn ?

hunshcn:
> Hm, I think we should probably have the same API as `buildifier_test` with a `no_sandbox` attribute https://github.com/bazelbuild/buildtools/pull/1092/files#diff-c5e2ba5db338e88cf7c850afd25c044dfa97198fefa4b296370c9286ab81d415R97 in addition to `srcs`/`workspace`, any thoughts about that @hunshcn ?

While providing workspace and set no_sandbox=True is strongly binding, this does provide clear semantics that I think are great.

alexeagle:
Yeah I think it's just there to ensure that you read the documentation and know that the test will never be a cache hit. Will add.

alexeagle:
Thanks for getting this together @hunshcn

alexeagle (approved):

Zemnmez:
Thanks for this!

PR #191, opened 2024-03-30
chore: test on Bazel 7

alexeagle:


alexeagle:
@gregmagolan I need to track down your recent changes to green this up - unless you'd like to?

MODULE.bazel
```diff
@@ -28,3 +28,12 @@ multitool = use_extension("@rules_multitool//multitool:extension.bzl", "multitoo
 multitool.hub(lockfile = "//format:multitool.lock.json")
 multitool.hub(lockfile = "//lint:multitool.lock.json")
 use_repo(multitool, "multitool")
+
+# Locally, use newer bazel-lib for stardoc
+single_version_override(
```
alexeagle, line 33: note, we're in RC already so I'm keeping safe from the semver 👮🏻 by allowing users to stay on bazel-lib 1 as minimum

gregmagolan (approved):

MODULE.bazel
gregmagolan: sure... we can save that rules rules_lint 2

PR #196, opened 2024-04-04
feat: shellcheck apply fixes

alexeagle:
### Type of change

- Relevant documentation has been updated? We don't document which linters have fixes auto-applied, but I'm not sure that we should, since it will be hard to keep that updated.

### Test plan

```
example% ./lint.sh --fix src:hello_shell
From bazel-out/darwin_arm64-fastbuild/bin/src/shellcheck.hello_shell.aspect_rules_lint.report:

In src/hello.sh line 3:
[ -z $THING ] && echo "hello world"
     ^----^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
[ -z "$THING" ] && echo "hello world"

For more information:
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...

patching file 'src/hello.sh'

example % git diff src  
diff --git a/example/src/hello.sh b/example/src/hello.sh
index 51afcbc..64b7956 100644
--- a/example/src/hello.sh
+++ b/example/src/hello.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-[ -z $THING ] && echo "hello world"
+[ -z "$THING" ] && echo "hello world"
 
 # Note, we should not get a lint here because the .shellcheckrc excludes it
 [ ! -z "$foo" ] && echo "foo"
```

### DIFF
--- a/docs/shellcheck.md
+++ b/docs/shellcheck.md
@@ -64,7 +64,7 @@ Attrs:
 ## shellcheck_action
 
 <pre>
-shellcheck_action(<a href="#shellcheck_action-ctx">ctx</a>, <a href="#shellcheck_action-executable">executable</a>, <a href="#shellcheck_action-srcs">srcs</a>, <a href="#shellcheck_action-config">config</a>, <a href="#shellcheck_action-report">report</a>, <a href="#shellcheck_action-use_exit_code">use_exit_code</a>)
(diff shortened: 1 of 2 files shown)


mrmeku (approved):
Let me know if you'd like some help in the manual testing department, but the change itself looks good

alexeagle:
You'll get a chance to try it when we dogfood in silo!

PR #203, opened 2024-04-09
refactor: bring linters under rules_multitool

alexeagle:
Continues the work from setting up formatters this way.

Includes an upgrade of shellcheck to the latest release.

### Type of change

    1. Remove any `fetch_shellcheck` or `fetch-golangci_lint` calls.
    2. Replace references of a local shellcheck binary with `@multitool//tools/shellcheck` and delete the local binary
    3. Replace references of golangci-lint binary with `@multitool//tools/golangci-lint` and delete local binary or alias

### Test plan

### DIFF
--- a/example/MODULE.bazel
+++ b/example/MODULE.bazel
@@ -1,7 +1,7 @@
 "Bazel dependencies"
 
 bazel_dep(name = "aspect_rules_lint", version = "0.0.0")
-bazel_dep(name = "aspect_bazel_lib", version = "2.5.0")
+bazel_dep(name = "aspect_bazel_lib", version = "2.6.1")
 bazel_dep(name = "aspect_rules_js", version = "1.40.0")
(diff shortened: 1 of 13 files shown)


example/MODULE.bazel
```diff
@@ -1,7 +1,7 @@
 "Bazel dependencies"
 
 bazel_dep(name = "aspect_rules_lint", version = "0.0.0")
-bazel_dep(name = "aspect_bazel_lib", version = "2.5.0")
+bazel_dep(name = "aspect_bazel_lib", version = "2.6.1")
```
gregmagolan, line 4: If you're bumping to latest then 2.7.0 is latest atm

gregmagolan (approved):
🌮

example/MODULE.bazel
alexeagle: Ah, I should have commented. This bump is minimal needed to fix the GLIBC version error I was getting on BCR releases.

PR #207, opened 2024-04-11
feat: remove golangci-lint

### LINKED ISSUE #129: [Bug]: golangci-lint doesn't pull in dependencies
### What happened?

golangci-lint doesn't just look at the .go file specified on the command-line, it also needs to pull in metadata about e.g. types from other files that are referenced from the file that is being linted. Without this, it complains about missing files.

The lint target should inherit all dependencies of the original target, as well as dependencies on any generated files in ``srcs`` as those are currently being filtered out (since they don't newed to be linted).

### Version

Output of `bazel --version`:

bazel 7.0.2

N/A, reproducible using example from rules_lint

Go / golang-ci

### How to reproduce

```shell
I'll attach a PR to this bug that demonstrates the issue with the example in the rules_lint repo.
```

### Any other information?

_No response_

alexeagle:
In preparation for 1.0, we want all linters to work correctly. In the discussion for #129 we decided that is a fatal bug, and no one is able to fix it in the short term.

We're happy to add Go linting back to rules_lint when someone has a correct implementation that handles transitive srcs. The code will still be here in the git history when it's time to revive it.

Closes #129

---

### Type of change

Well, the opposite of a feature, whatever you call that...

BREAKING: there is no longer a  `lint_golangci_aspect` symbol. Support for Go linting will hopefully be restored in a later release.

### Test plan


gregmagolan (approved):

alexeagle:
bummer!

PR #210, opened 2024-04-12
feat(format): Add ability to ignore files

mrmeku:
For any files that you want rules_lint to ignore you can now give them an attribute via .gitattributes files repository.

The following attributes are supported
- rules_lint_ignored
- gitlab-generated
- linguist-generated

---

### Type of change

### Test plan
- New test cases added

### DIFF
--- a/docs/linting.md
+++ b/docs/linting.md
@@ -92,17 +92,6 @@ By default, we filter out generated files from linting.
 
 To bypass this filter, add `tags=["lint-genfiles"]` to a target to force all the `srcs` to be linted.
 
-## Ignoring files explicitly
-
-Commonly, the underlying linters that rules_lint invokes provide their own methods of excluding files from linting (.prettierignore for example). At times when that is not the case, rules_lint provides its
-own escape hatch to exclude files from linting using attributes specified via [`.gitattributes` files](https://git-scm.com/docs/gitattributes).
-
-If any of following attributes are set on a file it will be excluded from linting:
-
-- `rules-lint-ignored`
-- `gitlab-generated`
-- `linguist-generated`
-
 ## Debugging
 
 Some linters honor the debug flag in this repo. To enable it, add a Bazel flag:
--- a/docs/linting.md
+++ b/docs/linting.md
@@ -92,6 +92,17 @@ By default, we filter out generated files from linting.
 
 To bypass this filter, add `tags=["lint-genfiles"]` to a target to force all the `srcs` to be linted.
 
+## Ignoring files explicitly
+
+Commonly, the underlying linters that rules_lint invokes provide their own methods of excluding files from linting (.prettierignore for example). At times when that is not the case, rules_lint provides its
+own escape hatch to exclude files from linting using attributes specified via [`.gitattributes` files](https://git-scm.com/docs/gitattributes).
+
+If any of following attributes are set on a file it will be excluded from linting:
+
+- `rules-lint-ignored`
+- `gitlab-generated`
+- `linguist-generated`
+
 ## Debugging
 
 Some linters honor the debug flag in this repo. To enable it, add a Bazel flag:
--- a/example/.gitattributes
+++ b/example/.gitattributes
@@ -1,0 +1,7 @@
+# Do not try and format generated code
+pnpm-lock.yaml linguist-generated
+
+# Any of the below attributes can be set to ignore a file
+src/subdir/rules-lint-ignored.js rules-lint-ignored
+src/subdir/gitlab-generated.js gitlab-generated
+src/subdir/linguist-generated.js linguist-generated
--- a/example/src/subdir/gitlab-generated.js
+++ b/example/src/subdir/gitlab-generated.js
@@ -1,0 +1,2 @@
+// Ignored by https://github.com/aspect-build/rules_lint/blob/example/.gitattributes
+console.log("console is not allowed, but this file is ignored");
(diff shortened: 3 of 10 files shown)


example/.gitattributes
```diff
@@ -0,0 +1 @@
+src/subdir/ignored.js rules_lint_ignored=true
```
alexeagle, line 1: could we also add a line like
`pnpm-lock.yaml linguist-generated=true`
and assert that we don't try to format generated code?

format/private/format.sh
```diff
@@ -94,9 +94,20 @@ function ls-files {
...
-          grep -vE "^$(git ls-files --deleted)$" || true;
-        }
+        files=$(git ls-files --cached --modified --other --exclude-standard "${patterns[@]}" "${patterns[@]/#/*/}" | {
+          grep -vE \
+            "^$(git ls-files --deleted)$" \
+          || true;
+        })
+        non_ignored_files=()
+        for file in "${files[@]}"; do
+            if [[ $(git check-attr -a -- $file | grep "rules_lint_ignored: true" ) == "" ]]; then
```
alexeagle, line 104: is this parse-brittle? What if my `.gitattributes` had no space like `rules_lint_ignored:true` ?

example/src/subdir/ignored.js
```diff
@@ -0,0 +1 @@
+console.log("This should be lint error");
```
alexeagle, line 1: Add a comment that we don't expect the error will be reported, because it appears in the .gitattributes (add the backlink)

format/private/format.sh
```diff
@@ -94,9 +94,20 @@ function ls-files {
...
-        git ls-files --cached --modified --other --exclude-standard "${patterns[@]}" "${patterns[@]/#/*/}" | {
-          grep -vE "^$(git ls-files --deleted)$" || true;
-        }
+        files=$(git ls-files --cached --modified --other --exclude-standard "${patterns[@]}" "${patterns[@]/#/*/}" | {
+          grep -vE \
+            "^$(git ls-files --deleted)$" \
+          || true;
+        })
+        non_ignored_files=()
+        for file in "${files[@]}"; do
```
alexeagle, line 103: instead of looping the files and running `git check-attr` N times with one argument each, is it possible to run `git check-attr` once with N arguments?

alexeagle (commented):
@thesayyn is our bash 🧙🏻‍♂️

alexeagle:
Also needs a doc update so users know that the feature exists and what syntax it supports

format/private/format.sh
mrmeku: It wasn't brittle before, but its more idiomatic now since I just check if the attribute is set or unset rather than for a specific value

format/private/format.sh
mrmeku: Done

docs/linting.md
```diff
@@ -92,6 +92,17 @@ By default, we filter out generated files from linting.
 
 To bypass this filter, add `tags=["lint-genfiles"]` to a target to force all the `srcs` to be linted.
 
+## Ignoring files explicitly
```
alexeagle, line 95: this is the wrong file, I think you want to edit `formatting.md`. This one has nothing to do with `format.sh`

example/.gitattributes
```diff
@@ -0,0 +1,7 @@
+# Do not try and format generated code
+pnpm-lock.yaml linguist-generated
+
+# Any of the below attributes can be set to ignore a file
+src/subdir/rules_lint_ignored.js rules_lint_ignored
```
alexeagle, line 5: nit: seems like the convention is "kebab case" so I don't think we should be different and use snake case, will be harder to remember what to type

example/src/subdir/gitlab-generated.js
```diff
@@ -0,0 +1,2 @@
+// Ignored by https://github.com/aspect-build/rules_lint/blob/example/.gitattributes
+console.log("This is an intentional lint error");
```
alexeagle, line 2: it's a format error, not lint right?

mrmeku: No, the console.log is made to be a lint error because `no-console` is enabled in .eslintrc

mrmeku: I've updated the string to make that more clear hopefully

example/.gitattributes
mrmeku: Done

docs/linting.md
mrmeku: Done

example/src/subdir/gitlab-generated.js
alexeagle: But the ignore handling you've added is only for the formatter, which is unrelated to eslint. So this can't possibly be testing the feature

alexeagle:
The comments are all resolved, but the changes here still seem wrong to me. The format and lint capabilities of this repo are totally orthogonal - format is just a binary we run, which you've added the ignore feature to. Linting is an aspect over the library graph and isn't affected by the new feature, yet you're using linting as the test fixture.

mrmeku:
@alexeagle should linting be affected by the new feature? The confusion here just comes from my lack of familiarity with the implementation details of the repo. I personally think that both formatting and linting should be affected. If you agree I'll implement it that way and make sure to add a test case that actually exercises that logic. I just got confused

alexeagle:
We could discuss that as a follow up... I doubt we want to do that since linters should see the same srcs that the shadowed *_library does.
Lemme know if you'd like some tour of the repo! Maybe Jesse and Derek too

alexeagle:
Ping @mrmeku I'm still hoping you'll fix this to not suggest that linting is affected by the ignore feature, at least for this PR

mrmeku:
Done!

alexeagle (approved):
nice!

PR #211, opened 2024-04-16
fix: pass flag to shfmt to ignore passed files in check mode

daewok:
---

Pass the `--apply-ignore` flag the shfmt when checking to see if files are already formatted.

### Type of change

shfmt will now respect ignores in .editorconfig even in check mode.

### Test plan

I didn't see any existing tests for check mode.


alexeagle:
Ah, I had to go track down https://github.com/mvdan/sh/issues/1037 to find the context.  Thanks!

alexeagle (approved):

PR #216, opened 2024-04-23
chore: add sh_test for ls-files function

alexeagle:
It caught an issue when no files are found, we cannot run git check-attr

---

### Type of change

### Test plan

- New test cases added


format/private/format.sh
```diff
@@ -13,6 +13,9 @@ source "${RUNFILES_DIR:-/dev/null}/$f" 2>/dev/null || \
   { echo>&2 "ERROR: runfiles.bash initializer cannot find $f. An executable rule may have forgotten to expose it in the runfiles, or the binary may require RUNFILES_DIR to be set."; exit 1; }; f=; set -e
 # --- end runfiles.bash initialization v3 ---
 
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+source "$SCRIPT_DIR/ls-files.sh"
```
mrmeku, line 17: source "$SCRIPT_DIR/private/ls-files.sh"

mrmeku (approved):
Looks good to me on the code. You'll need to update the directory you're sourcing from though to make CI pass

format/private/format.sh
alexeagle: This script is already in `/private` so that doesn't feel like the right fix... looking

alexeagle: There's not a good answer for the runfiles lookup that would be needed. Went back to a single format.sh script to avoid inventing something complicated.

PR #215, opened 2024-04-22
feat(lint): Add Kotlin linter using `ktlint`

smocherla-brex:
This PR adds support for a linter aspect with `bazel lint` or `aspect lint` with `ktlint` using https://github.com/pinterest/ktlint.

I followed the linter README and added support for `ktlint` here with some of the CLI attributes pulled in as the aspect attributes. I also updated the `example/` workspace to pull in the ktlint aspect and run it over a `kt_jvm_library` target from rules_kotlin. `ktlint` itself is to be installed in the user's workspace by using maven_install and then wrapping it with a `java_binary` since it relies on the Java toolchain.

Right now it supports a subset of the CLI options. Follow-up PRs can be done to support:
- A custom ruleset jar
- Customer reporters (checkstyle) and file formats

### Type of change

- New feature or functionality (change which adds functionality)

**For changes visible to end-users**
- Relevant documentation has been updated

### Test plan

- New test cases added (in `example/`)
- Manual testing; please provide instructions so we can reproduce:

I updated `example/` and also ran `lint.sh src:all`. Now it includes ktlint output for the new target added here

```
❯ ./lint.sh src:all
INFO: Invocation ID: 345cd573-bd9d-4197-8cd6-999689265a6c
WARNING: Build options --java_language_version, --java_runtime_version, and --tool_java_runtime_version have changed, discarding analysis cache (this can be expensive, see https://bazel.build/advanced/performance/iteration-speed).
INFO: Analyzed 15 targets (10 packages loaded, 17599 targets configured).
INFO: Found 15 targets...
INFO: Elapsed time: 0.591s, Critical Path: 0.24s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Build Event Protocol files produced successfully.
...
...
From bazel-out/k8-fastbuild/bin/src/ktlint.hello_kt.aspect_rules_lint.report:
src/hello.kt:1:1: File name 'hello.kt' should conform PascalCase (standard:filename)
src/hello.kt:1:1: Imports must be ordered in lexicographic order without any empty lines in-between with "java", "javax", "kotlin" and aliases in the end (standard:import-ordering)
src/hello.kt:1:1: Wildcard import (standard:no-wildcard-imports)
src/hello.kt:5:1: Unexpected indentation (2) (should be 4) (standard:indent)

Summary error count (descending) by rule:
  standard:filename: 1
  standard:import-ordering: 1
  standard:indent: 1
  standard:no-wildcard-imports: 1

...
...
[...]

PR #220, opened 2024-04-28
ktlint: Add option to use custom rulesets

smocherla-brex:
Follow-up on https://github.com/aspect-build/rules_lint/pull/215, this adds support for the `--ruleset` argument with `ktlint` https://pinterest.github.io/ktlint/1.2.1/install/cli/#rule-sets

For most people, the standard rulesets may suffice. However, in large codebases it's common to write your own custom rules and use them on top of standard rules. This adds support for it by updating the `ktlint` aspect.

### Test plan

This is not a breaking change and existing test cases should pass. I'd like to add testcases for this code path, but that'd require adding a custom ruleset dependency here. There's no popular OSS custom rulesets and not sure it has to live here (although I'd be happy to add). I've tested this against our custom rules and it seems to work as expected.

### DIFF
--- a/docs/ktlint.md
+++ b/docs/ktlint.md
@@ -29,6 +29,28 @@ ktlint = ktlint_aspect(
 )
 ```
 
+If you plan on using Ktlint [custom rulesets](https://pinterest.github.io/ktlint/1.2.1/install/cli/#rule-sets), you can also declare 
+an additional `ruleset_jar` attribute pointing to your custom ruleset jar like this
+
+```
+java_binary(
+    name = "my_ktlint_custom_ruleset",
+    ...
+)
+
+ktlint = ktlint_aspect(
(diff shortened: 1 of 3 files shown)


alexeagle (approved):
Nice! It would be good to illustrate the usage with an example as well, though it doesn't necessarily have to go in this PR.

smocherla-brex:
> Nice! It would be good to illustrate the usage with an example as well, though it doesn't necessarily have to go in this PR.

I updated the docs on how to use it. There don't seem to be any well-maintained custom OSS rulesets which can be used directly here. Alternative is to use a dummy/example ruleset in the `example/` workspace but that requires adding a bunch of ktlint dependencies to the maven lock file and build from source. I can add it in a separate PR if desired.

alexeagle (approved):

PR #221, opened 2024-04-30
chore: compatibility with rules_js 2

alexeagle:
### Test plan

- New test cases added

The example now tests rjs2 only with MODULE.bazel, and rjs1 with WORKSPACE.


alexeagle:
Flipped to draft until the rules_ts 3 beta is released

gregmagolan:
Updated to rules_ts 3.0.0-alpha.0

lint/eslint.bzl
```diff
@@ -63,13 +63,25 @@ def _gather_inputs(ctx, srcs):
...
+        # rules_js 1.x
+        js_inputs = js_lib_helpers.gather_files_from_js_providers(
+            ctx.attr._config_files + [ctx.attr._workaround_17660, ctx.attr._formatter],
+            include_transitive_sources = True,
+            include_declarations = False,
+            include_npm_linked_packages = True,
+        )
+    else:
+        # rules_js 2.x
+        js_inputs = js_lib_helpers.gather_files_from_js_infos(
```
gregmagolan, line 76: I don't expect this API to change between rules_js 2.0.0-alpha.4 and final so this PR is safe to land... tho maybe we should wait until RC so we don't have the public example point to alpha releases of rules_js & rules_ts

alexeagle: Gotta land soon so we can try rjs2 downstream

PR #225, opened 2024-05-08
feat(lint): plumb exit codes as outputs rather than dropping them

alexeagle:
This allows downstream tools to return a correct exit code based on 'errors were reported by at least one linter'

---

It's only slightly visible, in that `lint.sh` now prints exit codes as well:

```
rules_lint/example$ ./lint.sh src:unused_import
INFO: Analyzed target //src:unused_import (5 packages loaded, 492 targets configured).
INFO: Found 1 target...
INFO: Elapsed time: 0.309s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build Event Protocol files produced successfully.
From bazel-out/k8-fastbuild/bin/src/ruff.unused_import.aspect_rules_lint.report:
src/unused_import.py:13:8: F401 [*] `os` imported but unused
Found 1 error.
[*] 1 fixable with the `--fix` option.

From bazel-out/k8-fastbuild/bin/src/ruff.unused_import.aspect_rules_lint.exit_code:
1

From bazel-out/k8-fastbuild/bin/src/flake8.unused_import.aspect_rules_lint.report:
src/unused_import.py:13:1: F401 'os' imported but unused

From bazel-out/k8-fastbuild/bin/src/flake8.unused_import.aspect_rules_lint.exit_code:
1
```

### Test plan

### DIFF
--- a/lint/ktlint.bzl
+++ b/lint/ktlint.bzl
@@ -56,7 +56,7 @@ load("//lint/private:lint_aspect.bzl", "LintOptionsInfo", "filter_srcs", "report
 
 _MNEMONIC = "ktlint"
 
-def ktlint_action(ctx, executable, srcs, editorconfig, report, baseline_file, java_runtime, ruleset_jar = None, use_exit_code = False):
+def ktlint_action(ctx, executable, srcs, editorconfig, report, baseline_file, java_runtime, ruleset_jar = None, exit_code = None):
     """ Runs ktlint as build action in Bazel.
 
     Adapter for wrapping Bazel around
@@ -71,7 +71,8 @@ def ktlint_action(ctx, executable, srcs, editorconfig, report, baseline_file, ja
         baseline_file: The file object pointing to the baseline file used by ktlint.
         java_runtime: The Java Runtime configured for this build, pulled from the registered toolchain.
         ruleset_jar: An optional, custom ktlint ruleset jar.
-        use_exit_code: whether a non-zero exit code from ktlint process will result in a build failure.
+        exit_code: output file to write the exit code.
+            If None, then fail the build when ktlint exits non-zero.
     """
 
     args = ctx.actions.args()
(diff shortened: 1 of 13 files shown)


lint/ktlint.bzl
```diff
@@ -104,27 +105,21 @@ def ktlint_action(ctx, executable, srcs, editorconfig, report, baseline_file, ja
...
     else:
-        command = """
-            # This makes hermetic java available to ktlint executable
-            export PATH=$PATH:$JAVA_HOME/bin
-
-            # Don't fail ktlint and just report the violations
-            {ktlint} $@ >{report} || true
-            """.format(ktlint = executable.path, report = report.path)
+        # Run ktlint with arguments passed
+        command += "{ktlint} $@ && touch {stdout}"
```
gregmagolan, line 117: should be `{ktlint} $@ >{stdout} || true` ??

gregmagolan: oh; nvmind... I was reading it wrong

lint/lint_test.sh
```diff
@@ -12,10 +12,21 @@ source "${RUNFILES_DIR:-/dev/null}/$f" 2>/dev/null || \
...
         cat ${report_file}
         exit 1
   fi
-done
\ No newline at end of file
+}
+
+function assert_exit_code_zero() {
+  exit_code=$(cat $(rlocation "$1"))
+  diagnostics_file=$(rlocation "$2")
```
gregmagolan, line 25: is this the `report_file`? "diagnostics" is a term not used anywhere else

lint/shellcheck.bzl
```diff
@@ -62,9 +67,10 @@ def _shellcheck_aspect_impl(target, ctx):
     if ctx.rule.kind not in ["sh_binary", "sh_library"]:
         return []
 
-    patch, report, info = patch_and_report_files(_MNEMONIC, target, ctx)
-    shellcheck_action(ctx, ctx.executable._shellcheck, filter_srcs(ctx.rule), ctx.file._config_file, report, ctx.attr._options[LintOptionsInfo].fail_on_violation)
```
gregmagolan, line 66: what are `LintOptionsInfo` used for? can the `fail_on_violation` attribute be removed with this change?

alexeagle: No, we still need it - but the logic to access it was moved to the shared helper function in `lint/private/lint_aspect.bzl`. That's nice because now access is in the same file as the definition.

lint/lint_test.sh
alexeagle: changed "diagnostics" to "output" in the case where a report output group could be confused with a report file in that output group

lint/shellcheck.bzl
alexeagle: Another way to explain: we want actions to succeed, it's then up to whatever tool interprets the lint results to decide whether developers workflow should be halted.

lint/lint_test.bzl
```diff
@@ -37,22 +37,36 @@ flake8_test(
...
+        if f.path.endswith(".report"):
+            output = f
+        elif f.path.endswith(".exit_code"):
+            exit_code = f
+        else:
+            fail("rules_lint_report output group contains unrecognized file extension: ", f.path)
+    if output and exit_code:
+        return "assert_exit_code_zero '{}' '{}'".format(to_rlocation_path(ctx, exit_code), to_rlocation_path(ctx, output))
+    if output:
+        return "assert_diagnostics_empty '{}'".format(to_rlocation_path(ctx, output))
```
gregmagolan, line 54: should be `assert_output_empty` ?

lint/private/lint_aspect.bzl
```diff
@@ -13,9 +13,21 @@ fail_on_violation_flag = rule(
...
-    report = ctx.actions.declare_file("{}.{}.aspect_rules_lint.report".format(mnemonic, target.label.name))
-    return report, OutputGroupInfo(rules_lint_report = depset([report]))
+# buildifier: disable=function-docstring
+def report_files(mnemonic, target, ctx):
+    outfile = "{}.{}.aspect_rules_lint.{}"
+    report = ctx.actions.declare_file(outfile.format(mnemonic, target.label.name, "report"))
+    outs = [report]
+    if ctx.attr._options[LintOptionsInfo].fail_on_violation:
+        # Fail on violation means the exit code is reported to Bazel as the action result
+        exit_code = None
```
gregmagolan, line 23: would it be useful in the future to also capture the exit code with `fail_on_violation` is true so it can be reported as "Linter X failed with exit code Y"?

alexeagle: there's nothing to capture in that case because it's either zero, or the action failed to run so it produces no outputs

gregmagolan: right. makes sense

PR #227, opened 2024-05-08
refactor: produce report and exit code in the same action with fixes

alexeagle:
ruff and eslint both have the ability to produce the report of unfixable violations in the same execution that applies fixes.
We can use this to avoid having a second action run under `--fix`
For eslint in particular, this is a required performance improvement.

---

- Breaking change (forces users to change their own code or config): yes, slightly

To produce fixes, users must now pass the `--@aspect_rules_lint//lint:fix` flag.
The `lint.sh` helper script has been updated accordingly, and will need to be re-vendored into your repository.

### Test plan

```
$ ./lint.sh --fix --dry-run src:unused_import
INFO: Analyzed target //src:unused_import (0 packages loaded, 2 targets configured).
From bazel-out/k8-fastbuild/bin/src/ruff.unused_import.aspect_rules_lint.report:
src/unused_import.py:16:7: F521 `.format` call has invalid format string: Single '{' encountered in format string
Found 2 errors (1 fixed, 1 remaining).

From bazel-out/k8-fastbuild/bin/src/flake8.unused_import.aspect_rules_lint.report:
src/unused_import.py:13:1: F401 'os' imported but unused
src/unused_import.py:17:7: F521 '...'.format(...) has invalid format string: Single '{' encountered in format string

From bazel-out/k8-fastbuild/bin/src/ruff.unused_import.aspect_rules_lint.patch:
--- a/src/unused_import.py
+++ b/src/unused_import.py
@@ -10,7 +10,6 @@
 # src/unused_import.py:12:8: F401 [*] `os` imported but unused
 # Found 1 error.
 # [*] 1 potentially fixable with the --fix option.
-import os
 
 # Another lint violation, which is not auto-fixable.
 # When running with `--fix` this one should be reported and lint should exit 1.

```

```
$ ./lint.sh --fix --dry-run src:ts
INFO: Analyzed target //src:ts (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
From bazel-out/k8-fastbuild/bin/src/ESLint.ts.aspect_rules_lint.report:
src/file.ts: line 7, col 1, Error - Unnecessary try/catch wrapper. (no-useless-catch)

1 problem

From bazel-out/k8-fastbuild/bin/src/ESLint.ts.aspect_rules_lint.patch:
--- a/src/file.ts
+++ b/src/file.ts
@@ -1,5 +1,5 @@
 // this is a linting violation, and is auto-fixed under `--fix`
-const a: string = "a";
+const a = "a";
 console.log(a);
 
 // linting violation; not auto-fixable
```

### DIFF
--- a/docs/eslint.md
+++ b/docs/eslint.md
@@ -86,7 +86,7 @@ https://eslint.org/docs/latest/use/command-line-interface
 ## eslint_fix
 
 <pre>
-eslint_fix(<a href="#eslint_fix-ctx">ctx</a>, <a href="#eslint_fix-executable">executable</a>, <a href="#eslint_fix-srcs">srcs</a>, <a href="#eslint_fix-patch">patch</a>)
+eslint_fix(<a href="#eslint_fix-ctx">ctx</a>, <a href="#eslint_fix-executable">executable</a>, <a href="#eslint_fix-srcs">srcs</a>, <a href="#eslint_fix-patch">patch</a>, <a href="#eslint_fix-stdout">stdout</a>, <a href="#eslint_fix-exit_code">exit_code</a>)
 </pre>
 
 Create a Bazel Action that spawns eslint with --fix.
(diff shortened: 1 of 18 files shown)


gregmagolan (approved):
🌮

PR #229, opened 2024-05-09
feat(ruff): check with --quiet, upgrade and bring it to multitool

hunshcn:


lint/multitool.lock.json
```diff
@@ -35,5 +35,41 @@
...
       }
     ]
+  },
+  "ruff": {
+    "binaries": [
+      {
+        "kind": "archive",
+        "os": "macos",
+        "cpu": "x86_64",
+        "url": "https://github.com/astral-sh/ruff/releases/download/v0.4.3/ruff-0.4.3-x86_64-apple-darwin.tar.gz",
```
alexeagle, line 45: currently users can pick a version of ruff, I filed https://github.com/theoremlp/rules_multitool/issues/32 so we could continue to allow that while also using multitool lockfile

lint/ruff.bzl
```diff
@@ -90,7 +90,7 @@ def ruff_fix(ctx, executable, srcs, config, patch):
         output = patch_cfg,
         content = json.encode({
             "linter": executable._ruff.path,
-            "args": ["check", "--fix"] + [s.path for s in srcs],
+            "args": ["check", "--quiet", "--fix"] + [s.path for s in srcs],
```
alexeagle, line 93: doesn't this also throw away other output that we might want to show users, in the failure case? I think we only wanted to have `--quiet` on success

alexeagle (commented):
sorry to waste your time here! I ended up fixing this yesterday as part of warning vs errors

hunshcn:
Well, it's a good thing to solve it.

hunshcn:
> Other parts of this pr maybe still valuable, let's discuss on an issue?

What did you mean? About multitool? It need a plan. I prefer https://github.com/theoremlp/rules_multitool/issues/32#issuecomment-2085524640

PR #231, opened 2024-05-09
fix: lint.sh pass norun_validations

alexeagle:
Linting is generally orthogonal to other "validation" actions - users want to see lint reports even if some validation actions are in a failing state.
See https://github.com/aspect-build/rules_ts/pull/574#issuecomment-2073632879

---

### Test plan

Ran `lint.sh`.


example/lint.sh
```diff
@@ -25,6 +25,7 @@ filter='.namedSetOfFiles | values | .files[] | ((.pathPrefix | join("/")) + "/"
 # NB: perhaps --remote_download_toplevel is needed as well with remote execution?
 args=(
 	"--aspects=$(echo //tools/lint:linters.bzl%{buf,eslint,flake8,ktlint,pmd,ruff,shellcheck,vale} | tr ' ' ',')"
+	"--norun_validations"
```
gregmagolan, line 28: Perhaps common this flag with a link to https://github.com/aspect-build/rules_ts/pull/574#issuecomment-2073632879 so future reads won't have to use blame to find out why it is here?

gregmagolan (approved):
🌮

PR #234, opened 2024-05-10
fix(format): Skip files with skip-worktree bit set

### LINKED ISSUE #233: [Bug]: format_multirun does not work well with git sparse-checkout
### What happened?

When using git sparse-checkout, `format_multirun` doesn't filter out files that have [skip-worktree bit](https://git-scm.com/docs/git-sparse-checkout#_internalssparse_checkout), and that results in `no such file or directory` errors when a formatter tries to format a file.

### Version

Output of `bazel --version`: bazel 7.1.2

`WORKSPACE` or `MODULE.bazel` file: 0.19.0

Language(s) and/or frameworks involved: Python, shell, Java, Yaml, Go, protobuf, starlark

### How to reproduce

```shell
Execute a target of `format_multirun` in sparse-checkout git repository.
```

### Any other information?

https://github.com/aspect-build/rules_lint/blob/65b2e4dc6f2907a12ddee8b1471fb3b0e96302cf/format/private/format.sh#L98

This line could be extended as `files=$(git ls-files -t --cached --modified --other --exclude-standard "${patterns[@]}" "${patterns[@]/#/*/}" | grep -v ^S | cut -f2 -d' ' | {` to support filtering out files with skip-worktree bit set.

honnix:
As reported in #233

Fixes #233 

---

Without this change, users would have seen errors, so this is indeed something visible to users as a bug fix.

### Test plan

I only tested this locally.

- Automated tests

```starlark
format_multirun(
    name = "format",
    yaml = "@multitool//tools/yamlfmt",
    ... # everything else as False
)
```

Create a git repo with two folders: `foo`, `bar`, each with an yaml file `foo.yaml` and `bar.yaml`.

```shell
$ git sparse-checkout init
$ git sparse-checkout add foo
$ bazel run //path/to/the/format:target
```

Yaml formatter would complain file not found or similar without this fix.

### DIFF
--- a/format/private/format.sh
+++ b/format/private/format.sh
@@ -95,7 +95,7 @@ function ls-files {
 
         # TODO: determine which staged changes we should format; avoid formatting unstaged changes
         # TODO: try to format only modified regions of the file (where supported)
-        files=$(git ls-files --cached --modified --other --exclude-standard "${patterns[@]}" "${patterns[@]/#/*/}" | {
+        files=$(git ls-files --cached --modified --other --exclude-standard "${patterns[@]}" "${patterns[@]/#/*/}" | grep -v ^S | cut -f2 -d' ' | {
           grep -vE \
             "^$(git ls-files --deleted)$" \
           || true;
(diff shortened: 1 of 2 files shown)


alexeagle (commented):
Awesome! I've just recently written https://github.com/aspect-build/rules_lint/blob/main/format/test/ls-files_test.sh - do you mind adding an assertion there to cover this?

alexeagle:
Note for audit, Spotify org has signed the CLA

PR #132, opened 2024-01-31
Add rustfmt support

psalaberria002:
### Type of change
- New feature or functionality (change which adds functionality)

**For changes visible to end-users**
- Relevant documentation has been updated

### Test plan

- Manual testing; please provide instructions so we can reproduce: cd example && bazel run format

Trying to add rustfmt as a formatter. rules_rust already has a toolchain for rustfmt, so I am using that in the example.
However, the runfiles cannot be located since we change directory to the root of the repo before running the formatters. Is there a workaround for this?

Initializing rules_rust in the example folder can be improved, but I want to make it work first.

```
$ bazel run format
Formatting Rust with rustfmt...
/home/paulsmsm/.cache/bazel/_bazel_paulsmsm/573d3651f449593b84c17a4875272cea/execroot/_main/bazel-out/k8-opt-exec-ST-13d3ddad9198/bin/tools/rustfmt.rustfmt: line 2: external/rules_rust~0.38.0~rust~rustfmt_nightly-2023-12-28__x86_64-unknown-linux-gnu_tools/bin/rustfmt: No such file or directory
FAILED: A formatter tool exited with code 123
Try running 'bazel run @@//tools:format' to fix this.
Error: bazel exited with exit code: 123
```

When calling `bazel run //tools:rustfmt -- $(pwd)/src/hello.rs` works as expected.

PR #235, opened 2024-05-10
feat: add rustfmt

alexeagle:
Picking up #132 again

---

Rust code may now be formatted by rustfmt

### Test plan

- New test cases added


alexeagle:
cc @illicitonion - thanks for making the upstream change!

mattem (approved):

PR #236, opened 2024-05-10
Use explicit labels to allow using from other repositories

calebzulawski:
I use rules_lint via a macro in my own company rules--this allows it to work that way via bzlmod.

---

### Test plan

Create a macro containing `format_multirun` and `format_test` in one repository/module, and call that macro from the root module.


alexeagle (approved):
thanks

PR #238, opened 2024-05-11
feat: eslint and ruff should operate in quiet mode

ewhauser:
If `eslint` and `ruff` are not operating in quiet mode, they spit out success messages when all the linter rules pass. For a project with a many linter targets, this is superfluous and annoying.

---

- Users will no longer see stdout output from `eslint` or `ruff` when targets that pass lint

### Test plan

Running `bazel lint` in `examples` shows the behavior.


alexeagle:
I think the test failure is legit?

ewhauser:
I'll open individual PRs for these instead

PR #240, opened 2024-05-16
chore: Real time instead of user time when formatting

honnix:
It seems to be more useful to display real time than user time for formatting and format checking.

---

- Searched for relevant documentation and updated as needed: yes (the video/gif? is largely outdated, but I'm not sure how to update)
- Breaking change (forces users to change their own code or config): yes (if user parses the bazel output to collect statistics)

Formatting and format checking now display a different message.

For example, instead of `Formatted Python in 0m0.004s` where time is `user time`, it is now `Formatted Python in 0m0.024s seconds` where time is `real time/wall-clock`.

### Test plan

`bazel //tools:format` or `bazel //tools:format.check`

### DIFF
--- a/format/private/format.sh
+++ b/format/private/format.sh
@@ -131,34 +131,38 @@ function time-run {
   local bin="$1" && shift
   local lang="$1" && shift
   local silent="$1" && shift
-  local tuser
-  local tsys
+  local TIMEFORMAT="Formatted ${lang} in %lR"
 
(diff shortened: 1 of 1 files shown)


honnix:
@alexeagle Wdyt of this?

alexeagle (approved):

alexeagle:
SGTM thanks

PR #244, opened 2024-05-17
feat: format.sh has same behavior with arguments

mrmeku:
Previously, format.sh only took git attributes
into account if it was passed no arguments.

For both code paths if the last argument passed is `--disable_git_attribute_checks` filtering based on .gitattributes will be disabled.

---

### Test plan

- New test cases added

### DIFF
--- a/format/private/format.sh
+++ b/format/private/format.sh
@@ -49,6 +49,12 @@ trap on_exit EXIT
 # Exports a function that is similar to 'git ls-files'
 # ls-files <language> [<file>...]
 function ls-files {
+    disable_git_attribute_checks=false
+    if [[ "${!#}" == "--disable_git_attribute_checks" ]]; then
+      set -- "${@:1:$#-1}"
+      disable_git_attribute_checks=true
+    fi
+
     language="$1" && shift;
     # Copied file patterns from
     # https://github.com/github-linguist/linguist/blob/559a6426942abcae16b6d6b328147476432bf6cb/lib/linguist/languages.yml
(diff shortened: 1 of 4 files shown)


format/private/format.sh
```diff
@@ -122,7 +119,22 @@ function ls-files {
           fi
           find_args+=("-name" "${patterns[$i]}")
         done
-        find "$@" "${find_args[@]}"
+        files=$(find "$@" "${find_args[@]}")
+    fi
+
+    if [[ $disable_git_attribute_checks == true ]]; then
```
alexeagle, line 125: I imagine the use case for this flag is "I'm not using git" - would it be equivalent to just short-circuit here if there's no `.gitattributes` file or if `git` isn't on the PATH or `git` says "this folder isn't a repository"?

alexeagle (approved):
maybe check if `format_test` needs any updates?

format/private/format.sh
mrmeku: Its a bit tricky since there can be one .gitattributes file per directory in a source repository so checking the root isn't sufficient and there's also the no arguments case where there's currently no way to avoid the call to `.git`. If we want to support gracefully handling git not being installed we should do it in its own PR. I did think about it though...

alexeagle: that's okay, we don't even need to document this flag, it's just a break-glass. SGTM

format/private/format.sh
```diff
@@ -122,7 +119,22 @@ function ls-files {
...
+      echo $files
```
honnix, line 126: This doesn't seem to be right because the call site expects line-separated list.

format/private/format.sh
```diff
@@ -122,7 +119,22 @@ function ls-files {
...
+      return
+    fi
+
+    if [[ $files != "" ]]; then
```
honnix, line 130: This part of change seems to cause quite some performance degradation for a big repo. I'm doing more tests to understand why.

format/private/format.sh
```diff
@@ -122,7 +119,22 @@ function ls-files {
...
+        git_attributes=$(git check-attr rules-lint-ignored linguist-generated gitlab-generated -- $files)
```
honnix, line 131: What I could see is, if some of the attributes is not specified, git will spit out things like `/path/to/file rules-lint-ignored: unspecified`, which means for each file we might have up to 3 lines of output, which I believe will in turn slow down grep.

`git check-attr -a -- /path/to/file` outputs nothing if no attribute is specified.

format/private/format.sh
honnix: I think it is due to `echo "$files" | tr \\n \\0`

format/private/format.sh
honnix: I think it is due to `echo "$files" | tr \\n \\0`.

PR #251, opened 2024-05-22
docs(format): include API docs for each language/tool

alexeagle:
Also remove any default tools for languages, forcing users to explicitly list an attribute and a tool value for each.

---

### Test plan

CI


example/tools/format/BUILD.bazel
```diff
@@ -54,21 +54,26 @@ format_multirun(
     name = "format",
     cc = "@llvm_toolchain_llvm//:bin/clang-format",
     css = ":prettier",
+    go = "@aspect_rules_lint//format:gofumpt",
```
gregmagolan, line 57: could use `BUILTIN_TOOL_LABELS["go"]` instead?

gregmagolan: oh.. nvmind.. I see how this is support to work now :)

gregmagolan (approved):
🦖

format/defs.bzl
```diff
@@ -55,30 +63,50 @@ def _format_attr_factory(target_name, lang, toolname, tool_label, mode):
...
-    Also produces a target `[name].check` which does not edit files, rather it exits non-zero
-    if any sources require formatting.
+    This macro produces a target named `[name].check` which does not edit files,
+    rather it exits non-zero if any sources require formatting.
 
-    Tools are provided by default for some languages.
-    These come from the `@multitool` repo.
-    Under --enable_bzlmod, rules_lint creates this automatically.
-    WORKSPACE users will have to set this up manually. See the release install snippet for an example.
+    Not recommended: to check formatting with `bazel test`, use [format_test](#format_test) instead.
```
mattnworb, line 100: does this mean to say that `[name].check` is not recommended, or that `format_test` is not recommended? Its pretty ambiguous

alexeagle: thanks, I'll clarify it's the latter

PR #252, opened 2024-05-22
fix: move mnemonic to after label in report, exit_code and patch output filename

gregmagolan:
Pulling the mnemonic out of the filename is difficult with the current naming:

```
path/to/target/mnemonic.label/with/slashes.aspect_rules_lint.report
```

This PR changes it to

```
path/to/target/label/with/slashes.aspect_rules_lint.mnemonic.report
```

so downstream tools can assume the file ends with `.mnemonic.report`

(same for `.exit_code` and `.patch` suffixes)

### DIFF
--- a/example/test/lint_test.bats
+++ b/example/test/lint_test.bats
@@ -44,7 +44,7 @@ EOF
 	assert_success
 
 	# Check that we created a 'patch -p1' format file that fixes the ESLint violation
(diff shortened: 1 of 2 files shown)


alexeagle:
As long as there isn't anyone downstream parsing the existing format yet lgtm

alexeagle (approved):

PR #256, opened 2024-05-27
chore: show less noisy prettier in example

alexeagle:
---

### Test plan


gregmagolan (approved):

example/tools/format/BUILD.bazel
```diff
@@ -14,6 +14,7 @@ prettier.prettier_binary(
     name = "prettier",
     # Allow the binary to be run outside bazel
     env = {"BAZEL_BINDIR": "."},
+    fixed_args = ["--log-level=silent"],
```
psalaberria002, line 17: nice! you may want to set it to error for logging errors on failure. Haven't tested the difference, but silent might be too silent.

PR #259, opened 2024-05-29
Make linter aspect provider and helpers public

psalaberria002:
We have implemented a couple of linters in our private repository, and it would be nice if we could import the provider and the helpers without patching rules_lint on our end.

---

The functions would now be publicly exposed.

### DIFF
--- a/lint/BUILD.bazel
+++ b/lint/BUILD.bazel
@@ -187,6 +187,15 @@ bzl_library(
 )
 
 bzl_library(
+    name = "lint_aspect",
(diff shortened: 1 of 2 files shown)


psalaberria002:
> I think I'm not quite ready to make these public API yet. You don't need to patch rules_lint to use these right? Just have to ignore a warning from buildifier that you're `load`ing from a private API?

fair enough. yes, the buildifier check is the only blocker.

alexeagle:
Just pass --nocheck_visibility when testing/updating docs? Visibility is an optional feature

PR #262, opened 2024-05-29
fix(format): "File name too long" when using --disable_git_attribute_checks

### LINKED ISSUE #261: [Bug]: "File name too long" when using --disable_git_attribute_checks with format
### What happened?

We've noticed that running `format_multirun` on a codebase with ~1500 .java files (about ~129k LOC) is much slower since upgrading to rules_lint 1.0.0-rc1.

As a part of trying to understand what changed, we thought that the work that format.sh does to check git attributes on each file (https://github.com/aspect-build/rules_lint/pull/244 seems related) might have an impact, and found that running

```
bazel run format
```

causes the script to exit with a line like:

```
... com/google/thirdparty/publicsuffix/TrieParser.java guava/refactorings/CharMatcherRewrite.java guava/refactorings/TraverserRewrite.java: File name too long
```

### Version

Output of `bazel --version`: bazel 7.2.0rc1

```
bazel_dep(name = "aspect_rules_lint", version = "1.0.0-rc1")
```

n/a

### How to reproduce

```shell
clone https://github.com/mattnworb/aspect-lint-rc1-repro

add a ton of Java files, for example with:

git subtree add --prefix guava git@github.com:google/guava.git master --squash

bazel run format -- --disable_git_attribute_checks
```

mattnworb:
this fixes #261, which is caused by the branch in this code when `disable_git_attribute_checks` is set returning the full list of files rather than echoing them one at a time as the other branch does:

https://github.com/aspect-build/rules_lint/blob/30c5c399b07f0327b9d4ffaf67370d7be7e79285/format/private/format.sh#L129-L142

### Test plan

- incorrect existing test fixed

I've created <https://github.com/mattnworb/aspect-lint-rc1-repro> to reproduce #261, and the branch <https://github.com/mattnworb/aspect-lint-rc1-repro/tree/fix-disable_git_attribute_checks> tests this patch.

It would be a good idea to add a test case to this repo for this sort of thing, to prevent it from happening in the first place, but I have not yet done this.

### DIFF
--- a/format/test/ls-files_test.sh
+++ b/format/test/ls-files_test.sh
@@ -2,8 +2,7 @@
 
 set -o nounset -o errexit -o pipefail
 
-BUILD_WORKSPACE_DIRECTORY=$TEST_TMPDIR
-source $TEST_SRCDIR/_main/format/private/format.sh
+BUILD_WORKSPACE_DIRECTORY=$TEST_TMPDIR source "$TEST_SRCDIR/_main/format/private/format.sh"
 
 git init --initial-branch=test
 
@@ -51,7 +50,11 @@ js=$(ls-files JavaScript src.js gen1.js gen2.js gen3.js)
     exit 1
 }
 js=$(ls-files JavaScript src.js gen1.js gen2.js gen3.js --disable_git_attribute_checks)
-[[ "$js" == "src.js gen1.js gen2.js gen3.js" ]] || {
+expected='src.js
+gen1.js
+gen2.js
+gen3.js'
+[[ "$js" == "$expected" ]] || {
     echo >&2 -e "expected ls-files to return src.js gen1.js gen2.js gen3.js, was\n$js"
     exit 1
 }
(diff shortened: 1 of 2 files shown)


format/test/ls-files_test.sh
```diff
@@ -51,7 +50,11 @@ js=$(ls-files JavaScript src.js gen1.js gen2.js gen3.js)
     exit 1
 }
 js=$(ls-files JavaScript src.js gen1.js gen2.js gen3.js --disable_git_attribute_checks)
-[[ "$js" == "src.js gen1.js gen2.js gen3.js" ]] || {
+expected='src.js
+gen1.js
+gen2.js
+gen3.js'
+[[ "$js" == "$expected" ]] || {
```
mattnworb, line 57: note that lower down in this file there is (existing) code in this structure, and expecting the return value to be spread over several liens

alexeagle (approved):
awesome, thanks a lot for fixing this!

PR #263, opened 2024-05-29
feat: add format_multirun() attribute to disable slow gitattribute checks

mattnworb:
We recently upgraded a repo containing ~1500 Java source files from `0.21.0` to `v1.0.0-rc1` and noticed that running `format_multirun()` over the whole codebase without any arguments (e.g. `bazel run format`) was significantly slower than before.

For example, we would see execution times like

```
> time bazel run //:format
...
Formatted Python in 0m0.062s
Formatted Starlark in 0m0.087s
Formatted Java in 0m3.535s
Formatted Go in 0m0.039s
Formatted Shell in 0m0.051s
Formatted Protocol Buffer in 0m1.878s
Formatted YAML in 0m0.154s

real    0m22.364s
user    0m41.983s
sys     0m6.982s
```

note that these numbers don't add up - in particular, a large amount of the 22s `real` time was waiting for the Java formatter.
 
Before the upgrade, we'd see Java formatter times like:

```
> time bazel run //:format_Java_with_java-format
...
Formatted Java in 0m3.195s

real    0m6.734s
user    0m25.887s
sys     0m2.578s
```

Investigating a bit more, we traced it down to the addition to `format/private/format.sh` to check the git attributes of each and every file (when the script is run with no arguments / specific source files) to find which files should not be formatted, which was added in #210 and

The git attributes check can be skipped by passing `--disable_git_attribute_checks` as an argument to the format target, but we don't want to have to instruct every engineer working in this repo to set this (or set it always in everyone's pre-commit hook, CI, etc.). Note that this flag is actually broken at the moment for large repos, which is fixed in #262.

This commits adds the ability to always disable these checks by setting an attribute on `format_multirun`, for anyone who does not have any use case to ignore individual files via gitattributes.

Arguably the default behavior should be reversed, since checking the attributes of each file individually does not scale well, or perhaps there are more performant ways to ignore formatting in specific files (such as listing files in an attr on `format_multirun`?).

    ```
     added `disable_git_attributes_check` to `format_multirun()` to disable checking the git attributes of each file-to-be-formatted to see if the formatting the file should be skipped, which can be excessively slow with a large number of files
    ```

### Test plan

tested in <https://github.com/mattnworb/aspect-lint-rc1-repro/tree/attribute-for-disable_git_attribute_checks>, any advice on how to add tests here for this?

### DIFF
--- a/docs/format.md
+++ b/docs/format.md
@@ -52,7 +52,7 @@ languages(<a href="#languages-name">name</a>, <a href="#languages-cc">cc</a>, <a
 </pre>
 
 Language attributes that may be passed to [format_multirun](#format_multirun) or [format_test](#format_test).
-    
+
 Files with matching extensions from [GitHub Linguist] will be formatted for the given language.
 
 Some languages have dialects:
@@ -96,7 +96,7 @@ Some languages have dialects:
 ## format_multirun
 
 <pre>
-format_multirun(<a href="#format_multirun-name">name</a>, <a href="#format_multirun-jobs">jobs</a>, <a href="#format_multirun-print_command">print_command</a>, <a href="#format_multirun-kwargs">kwargs</a>)
+format_multirun(<a href="#format_multirun-name">name</a>, <a href="#format_multirun-jobs">jobs</a>, <a href="#format_multirun-print_command">print_command</a>, <a href="#format_multirun-disable_git_attribute_checks">disable_git_attribute_checks</a>, <a href="#format_multirun-kwargs">kwargs</a>)
 </pre>
 
 Create a [multirun] binary for the given languages.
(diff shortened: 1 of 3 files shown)


mattnworb:
Forgot to mention: I have not added a test here, although I'd be happy to if someone could give some advice on the best way to test something like this.

I've also created a reproducer for the slow git-attribute-check behavior in https://github.com/mattnworb/aspect-lint-rc1-repro - check out the `fix-disable_git_attribute_checks` branch (which has the fix in #262 patched into it), which on my Macbook shows:

```
> time bazel run format && time bazel run format --disable_git_attribute_checks
...
INFO: Running command line: bazel-bin/format.bash
Formatted Java in 0m10.065s

real    1m16.431s
user    2m26.079s
sys     0m9.715s

> time bazel run format -- --disable_git_attribute_checks
...
INFO: Running command line: bazel-bin/format.bash --disable_git_attribute_checks
Formatted Java in 0m14.142s

real    0m14.743s
user    2m7.299s
sys     0m2.402s
```

in that repo I cloned the source code for Google's Guava library, which has 3220 `.java` files - so checking the git attributes of 3220 files adds ~60 seconds to the run time.

alexeagle:
FYI @mrmeku 

Yeah I think we ought to be able to avoid the `O(n)` operation of looping over files to check which are known by git to be generated. Thanks for the reproduction repo - looks like a separate project of optimizing that code.

alexeagle (approved):

mrmeku:
If this was any other language than bash it would be so much easier :p. I'm sure there must be some way to make a hashmap in bash and I'll find it

alexeagle:
Writing a helper in another language is a valid option. We have the technology to distribute it as static linked binary toolchain!

alexeagle:
But it is a big unplanned undertaking so I think we should try to avoid for now. Could happen after 1.0

mrmeku:
https://github.com/aspect-build/rules_lint/pull/270/files

mrmeku:
Before https://github.com/aspect-build/rules_lint/pull/270 I tried to time it but after 3 minutes I got bored of waiting.

After https://github.com/aspect-build/rules_lint/pull/270/files
```
> time bazel run format -- --disable_git_attribute_checks
...
INFO: Running command line: bazel-bin/format.bash --disable_git_attribute_checks
Formatted Java in 0m7.727s

real    0m11.015s
user    1m23.526s
sys     0m1.871s
```

#profit

PR #270, opened 2024-05-31
feat: Optimize the git ignore checks

mrmeku:
Now rather than calling grep N times where N is
the number of files to be formatted, we call grep
0 times and simply iterate through the array of
lines outputted by git check-attr

---

### Test plan

Git attribute based ignoring files performance increase!

### DIFF
--- a/format/private/format.sh
+++ b/format/private/format.sh
@@ -135,14 +135,35 @@ function ls-files {
     fi
 
     if [[ $files != "" ]]; then
-        git_attributes=$(git check-attr rules-lint-ignored linguist-generated gitlab-generated -- $files)
-        for file in $files; do
-            # Check if any of the attributes we ignore are set for this file.
-            if ! grep -qE "(^| )$file: (rules-lint-ignored|linguist-generated|gitlab-generated): set($| )" <<< $git_attributes; then
-                echo "$file"
-            fi
-        done
-    fi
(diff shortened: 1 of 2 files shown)


alexeagle (approved):

alexeagle:
Would be interesting to repeat the benchmark using that reproduction posted on the original PR that described this issue

format/private/format.sh
```diff
@@ -135,14 +135,35 @@ function ls-files {
...
-        done
-    fi
+      git_attributes=$(git check-attr rules-lint-ignored linguist-generated gitlab-generated --stdin <<<"$files")
+
+      # Iterate over each line of the output, files will be reported multiple times, once per attribute checked. To keep from formatting a file twice, we keep track of when the file has changed.
+      last_file=""
+      attribute_set=false
+      while IFS= read -r line; do
+          # Extract the file name and attribute values
+          file=$(echo "$line" | cut -d ':' -f 1)
```
mattnworb, line 145: I think you can avoid having to call echo and cut for each line with Bash's `readarray` to split on a delimiter, like in Method 3 example 2 [here](https://www.geeksforgeeks.org/bash-scripting-split-string/)

mrmeku: Changed to `file="${line%%:*}"`

PR #273, opened 2024-06-03
Only run aspect actions when input files provided

psalaberria002:
Many tools fail if we don't provide any input files. Skipping the actions seems like the cleanest solution.

The other alternative is to always generate an empty report and a status code file with 0, but that seems unnecessary. On the other hand, it would make the tests cleaner. 

@alexeagle what do you think?

---

### Test plan

I am planning to update the tests in lint_tests, but I wanna make sure this is the right approach first.

### DIFF
--- a/docs/ktlint.md
+++ b/docs/ktlint.md
@@ -29,7 +29,7 @@ ktlint = ktlint_aspect(
 )
 ```
 
-If you plan on using Ktlint [custom rulesets](https://pinterest.github.io/ktlint/1.2.1/install/cli/#rule-sets), you can also declare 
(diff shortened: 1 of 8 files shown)


alexeagle:
I think this is the right start - but we have to create the expected empty file and file containing `0` - I think that's what the failing tests will say too

psalaberria002:
Lots of codepaths in aspects like the shellcheck lint, but it seems reasonable to me.
I tried to call the dummy action instead of any action that was called previously.

alexeagle (approved):
Nice! Thank you!! 🙇🏻

Issue #255, opened 2024-05-24
[Bug]: format_test has error FIX_TARGET


hunshcn:
### What happened?

error in

https://github.com/aspect-build/rules_lint/blob/d420ca0bdcaee949cb68a101de2a047d6733cf9c/format/defs.bzl#L48-L57

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: 7.1.1

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: MODULE.bazel

Language(s) and/or frameworks involved:

### Any other information?

Should allow format_test to customize FIX_TARGET.

---

Another idea is to merge format_test and format.check. The functions they undertake are similar.

But this is not very mature, and there are still some problems.
1. format_test is a test_suits, so there is no way to directly bazel run :format_test. This is the biggest difference.
2. format_test has srcs attribute, but format does not (usually it should be triggered from the command line).

@alexeagle WDYT

hunshcn:
@alexeagle

PR #280, opened 2024-06-11
fix(format): print FIX_CMD only when check

hunshcn:
fix https://github.com/aspect-build/rules_lint/issues/255


alexeagle (approved):
cool thanks

PR #281, opened 2024-06-12
feat: auto-update multitool versions

hunshcn:
### Test plan
- Manual testing


alexeagle (approved):
Cool! That's pretty simple :)

alexeagle:
Something is wrong, see https://github.com/aspect-build/rules_lint/pull/288 - these two are fighting with each other. The tools lockfile commit is placed in a PR, then the next step immediately overwrites that by force-pushing a commit that reverts it. I think they need to run in the same job together - will try that...

PR #285, opened 2024-06-14
feat: Add Cuda format support

jsharpe:
---

Adds support for formatting cuda files by passing clang_format target to the cuda attribute of format_multirun.

### Test plan

- New test cases added


alexeagle (approved):
nice, thanks!

PR #299, opened 2024-06-20
Update clang_tidy_wrapper.bash shebang

jsharpe:
Use env to find the path to bash - don't assume bash is located in /usr/bin


alexeagle (approved):
sorry I missed this in review!

PR #307, opened 2024-07-02
feat: parameterise clang-tidy aspect

peakschris:
Parameterize clang-tidy aspect name and mnemonic. We have a custom clang-tidy extension, and with this change we can create an extra aspect that runs it, and also have a default aspect that runs the OOTB clang-tidy.

---

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -7,6 +7,10 @@
 # https://bazelbuild.slack.com/archives/C014RARENH0/p1691158021917459?thread_ts=1691156601.420349&cid=C014RARENH0
 common --check_direct_dependencies=off
 
+# Provide more output on error in CI
+common --verbose_failures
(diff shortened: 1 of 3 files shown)


peakschris:
This is only supported for clang-tidy because I'm not aware of any other linters that can be extended codefully to support new rule checks (but I may be wrong, please let me know). Unless there are use cases to support parameterisation of the other ones, I'd prefer to keep this limited.

alexeagle:
Other linters have plugins, eslint for example could have different linter binaries setup with plugins specific to some subfolder of the repo. So the user might need to create more than one eslint aspect. Is that the use case here too?

alexeagle:
> Unless there are use cases to support parameterisation of the other ones

I think the conclusion here is that there are.

PR #314, opened 2024-07-05
Add comment ignores to example

SpencerC:
buf's [`allow_comment_ignores` functionality](https://buf.build/docs/configuration/v1/buf-yaml#allow_comment_ignores) does not work out of the box. It requires the bazel `--experimental_proto_descriptor_sets_include_source_info` to work. See: https://github.com/bufbuild/rules_buf/issues/64#issuecomment-2125324929

Many of these changes mirror https://github.com/bufbuild/rules_buf/pull/80, and I suspect adding them here as well will save a lot of folks needing to dive a layer down to figure out how to configure buf correctly.

---

### Test plan

- Manual testing; please provide instructions so we can reproduce: I'm guessing the behavior of the `example` directory is tested manually. To test, I suggest modifying the items in `example/buf.yaml` and the comment ignores in `example/src/file.proto` and verifying they behave as expected.


example/MODULE.bazel
```diff
@@ -99,6 +99,7 @@ use_repo(
 )
 
 buf = use_extension("@rules_buf//buf:extensions.bzl", "buf")
+buf.toolchains(version = "v1.31.0")
```
alexeagle, line 102: why add this line? Changing our example also implies that users have to change something, but we want to avoid breaking changes during RC

example/buf.yaml
```diff
@@ -1,4 +1,9 @@
 version: v1
+
 lint:
   use:
+    - DEFAULT
+  allow_comment_ignores: true
```
alexeagle, line 6: Yeah feels like everyone wants this feature :)

example/lint.sh
```diff
@@ -46,6 +46,9 @@ args+=(
 	# See https://github.com/aspect-build/rules_ts/pull/574#issuecomment-2073632879
 	"--norun_validations"
 	"--build_event_json_file=$buildevents"
+	# Required for the buf allow_comment_ignores option to work properly
+	# See https://github.com/bufbuild/rules_buf/issues/64#issuecomment-2125324929
+  "--experimental_proto_descriptor_sets_include_source_info"
```
alexeagle, line 51: @comius do you happen to know if anyone is working on making this flag non-experimental?

alexeagle (commented):
neat thanks!

FYI @perezd - I suspect a lot of Bazelers will use Buf via this ruleset, and maybe you'd like to reduce the ambition of rules_buf if you can avoid maintaining formatting&linting support

example/MODULE.bazel
SpencerC: I had two reasons I thought were good, but just double checked and found that it's not needed to achieve the core purpose of this PR. 😝 However, the changes between the default `buf` version (v1.28.1) and this version do not appear to be breaking, and it might behoove this project to encourage users to manage the underlying tool version separately from the Bazel rule version; you don't have to wait for `rules_buf` to update to get the latest features of `buf` itself, etc.

example/buf.yaml
SpencerC: That's the goal! And it does work without making an changes to `rules_lint` directly, just adding that flag does it.

example/MODULE.bazel
SpencerC: Happy to see it stay or go either way, LMK.

perezd:
Yeah, we would like to not maintain that. Sounds good to us!

example/MODULE.bazel
alexeagle: as long as it's not required for users, sure it's probably better for the example to be explicit so it's obvious to a user what version of buf they have and how to update it.

alexeagle (approved):

example/MODULE.bazel
alexeagle: it seems like your choice here is accidentally bringing some breaking changes to the output format, which is making it hard to green up the integration test. I'd suggest breaking it into two PRs doing one-thing-at-a-time so it's more straightforward to diagnose cause-and-effect of the test failure.

example/lint.sh
liningpan: This was the original issue added the experimental flag, and seems like it was automatically closed last year. https://github.com/bazelbuild/bazel/issues/9337

alexeagle (approved):

PR #321, opened 2024-07-09
feat: linter aspect factories allow customizing rule kinds

alexeagle:
This lets users adapt for situations where they have a custom rule, for example.

### DIFF
--- a/lint/eslint.bzl
+++ b/lint/eslint.bzl
@@ -55,7 +55,7 @@ See the [react example](https://github.com/bazelbuild/examples/blob/b498bb106b20
 
 load("@aspect_bazel_lib//lib:copy_to_bin.bzl", "COPY_FILE_TO_BIN_TOOLCHAINS", "copy_files_to_bin_actions")
 load("@aspect_rules_js//js:libs.bzl", "js_lib_helpers")
-load("//lint/private:lint_aspect.bzl", "LintOptionsInfo", "dummy_successful_lint_action", "filter_srcs", "patch_and_report_files", "report_files")
+load("//lint/private:lint_aspect.bzl", "LintOptionsInfo", "dummy_successful_lint_action", "filter_srcs", "patch_and_report_files", "report_files", "should_visit")
 
 _MNEMONIC = "AspectRulesLintESLint"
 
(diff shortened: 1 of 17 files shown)


alexeagle:
Note, I didn't touch `clang_tidy.bzl` in this PR because it uses the built-in `CcInfo` provider to determine what to visit.

lint/eslint.bzl
```diff
@@ -247,6 +248,9 @@ def lint_eslint_aspect(binary, configs):
                 allow_single_file = True,
                 cfg = "exec",
             ),
+            "_rule_kinds": attr.string_list(
```
sallustfire, line 251: @alexeagle The bazel documentation indicates that implicit attrs must be labels. Is that documentation out of date or in error?

https://bazel.build/rules/lib/globals/bzl.html#aspect

jbedard:
Are "kinds" purely based on the string name of a target? It doesn't matter where it loaded from, macro vs rule etc?

alexeagle:
@jbedard I believe the kind is what appears after the analysis phase, so macros have been expanded already and the `kind` is always what you would see in `bazel query`, which is to say it's the actual rule() declaration.

But yes, if it's named `ts_library` then this matches, whether it's the one from rules_ts or something the user defined.
We could decide to load provider symbols here instead, and then require that the rule provides something. For example ruff and flake8 could be based on a PyInfo provider. But of course we want a provider of sources, and adding a dependency just for that `load` is probably not desirable

lint/eslint.bzl
alexeagle: Interesting, I don't recall reading that doc. Apparently it's incorrect - I verified this has the desired behavior.

The real bug is there's no tech writer assigned to the Bazel project. I've been working on fixing it 😉

sallustfire: Good thing you tackled it then. I was thinking that restriction preventing passing the string array via context was going to make the implementation messier than I would have liked.

jbedard (approved):

lint/eslint.bzl
alexeagle: I always start from wishful thinking and then backoff in the face of errors :)

Issue #319, opened 2024-07-09
[FR]: Format GraphQL Files


sallustfire:
### What is the current behavior?

GraphQL files are ignored. Though GraphQL embedded in Javascript and Typescript sources will automatically be formatted.

### Describe the feature

It is common for Javascript/Typescript projects to be integrated with GraphQL APIs. Accordingly, Prettier ships with support for formatting GraphQL embedded inside Javascript/Typescript code as well as standalone GraphQL files.

`format_multirun` and should be extended to include GraphQL as a language and allow matching files to be formatted.

PR #322, opened 2024-07-09
feat: add GraphQL format support (#319)

sallustfire:
Prettier supports GraphQL natively. The file extensions are registered with the formatter and a basic format test is added.

---

- Suggested release notes appear below: 

GraphQL is added as a language to `format_multirun` to opt into formatting with prettier.

### Test plan

- New test cases added

An example hello.graphql file was added which is fixed when running `bazel run format` from `examples/`. There is also a unit test that prettier is invoked with graphql files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,11 +22,12 @@ Features:
 New tools are being added frequently, so check this page again!
 
 | Language               | Formatter             | Linter(s)        |
-| ---------------------- | --------------------- | ---------------- |
+|------------------------| --------------------- |------------------|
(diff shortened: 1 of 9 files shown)


sallustfire:
@alexeagle No rush on my part, but figured I'd flag you in case you hadn't seen this.

sallustfire:
@alexeagle No rush on my part, but figured I'd flag you in case you hadn't seen this.

sallustfire:
@alexeagle No rush on my part, but figured I'd flag you in case you hadn't seen this.

PR #337, opened 2024-07-17
fix(eslint): print relative paths in stylish output

alexeagle:
Thanks @gregmagolan for walking me through the technique required.

---

### Test plan

```
% ./lint.sh src:ts      
From bazel-out/darwin_arm64-fastbuild/bin/src/ts.AspectRulesLintESLint.out:

src/file.ts
  2:7  error  Type string trivially inferred from a string literal, remove type annotation  @typescript-eslint/no-inferrable-types
  7:1  error  Unnecessary try/catch wrapper                                                 no-useless-catch

✖ 2 problems (2 errors, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

### DIFF
--- a/lint/eslint.stylish-formatter.js
+++ b/lint/eslint.stylish-formatter.js
@@ -17,10 +17,13 @@
 // That means we can grab the path of the eslint entry point, which is beneath its node modules tree.
 const eslintEntry = process.argv[1];
 // Walk up the tree to the location where eslint normally roots the searchPath of its require() calls
-const searchPath = eslintEntry.substring(
-  0,
-  eslintEntry.length - "/node_modules/eslint/bin/eslint.js".length
-);
+const idx = eslintEntry.lastIndexOf("node_modules");
+if (idx < 0) {
+  throw new Error(
+    "node_modules not found in eslint entry point " + eslintEntry
+  );
+}
+const searchPath = eslintEntry.substring(0, idx);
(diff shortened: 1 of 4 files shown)


lint/eslint.stylish-formatter.js
```diff
@@ -0,0 +1,141 @@
...
+ * The three eslint dependencies should be loaded from the user's node_modules tree, not from rules_lint.
+ */
+
+// This script is used as a command-line flag to eslint, so the command line is "node eslint.js --format this_script.js"
+// That means we can grab the path of the eslint entry point, which is beneath its node modules tree.
+const eslintEntry = process.argv[1];
+// Walk up the tree to the location where eslint normally roots the searchPath of its require() calls
+const searchPath = eslintEntry.substring(
+  0,
+  eslintEntry.length - "/node_modules/eslint/bin/eslint.js".length
```
gregmagolan, line 22: It might be that some versions of eslint don't have the entry point as `bin/eslint.js` and this would fail. Most likely it is consistent but not necessarily going to stay that way. Safest way to proceed is probably to find the last `/node_modules/` in the path and strip to that point. If there is no `/node_modules/` segment then just error out.

gregmagolan (approved):
🌮

lint/eslint.stylish-formatter.js
alexeagle: Yeah I was thinking that for a moment, wow I'm lazy this week 🌵

PR #339, opened 2024-07-18
fix: include transitive type declarations with eslint

### LINKED ISSUE #335: [Bug]: Include transitive declarations with eslint inputs
### What happened?

The lint_eslint_aspect does not include type information and reports errors about the misuse of any when the typescript-eslint plugin rules tseslint.configs.recommendedTypeChecked are enabled.

### Version

Development (host) and target OS/architectures: Pop!_OS 22.04 LTS

Output of `bazel --version`: 7.1.1

### How to reproduce

_No response_

### Any other information?

While `gather_files_from_js_infos` is used, the sources are retrieved from the target as files and not JsInfo, so no transitive dependencies are inclluded. Fixing this requires getting the JsInfo provide from the target in the aspect implementation.

sallustfire:
typescript-eslint requires type declarations for imports when evaluating type checked lint rules. These declarations are included as inputs to the eslint action and a regression test under examples/ is included to confirm that types from both npm packages and locally imported sources are resolved.

### Test plan

- New test cases added

Fixes #335

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -1,5 +1,6 @@
 {
   "dependencies": {
+    "dayjs": "^1.11.11",
     "moment": "*"
   },
   "devDependencies": {
--- a/example/pnpm-lock.yaml
+++ b/example/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: "6.0"
+lockfileVersion: '6.0'
 
 settings:
   autoInstallPeers: true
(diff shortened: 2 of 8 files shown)


sallustfire:
@alexeagle A couple notes:
- A very old version of moment is included in the examples, so old that it doesn't even have type declarations. I opted to use dayjs as a placeholder to depend on the types of some npm package. I don't follow the thinking in subdir/ where moment is used, so there may be artifacts or this all can be consolidated.
- I did get bats to work, but ktlint was consistently failing. It looked like a structural error, so I commented it out, but we should revert that if it works in CI and there is just some hermeticity leak that causes it to fail on my system.
- There's a large diff in the pnpm lockfile, maybe someone applied a formatter to it? I believe this is the generated pnpm style though.

alexeagle (approved):
That was simple, thanks!

sallustfire:
@alexeagle No problem. Thanks for reviewing. It probably can be polished a little bit more e.g. we don't need sources for the deps just the transitive declarations. Stuff like that might help in larger repos?

PR #349, opened 2024-07-24
Add C example and file extension.

agoessling:
This adds `clang-format` support for C files `*.c`.  `clang-tidy` support already works correctly for `cc_*` targets with C sources.

It wasn't clear the best way to add `*.c` to `format.sh` given how it seems the output of `mirror_linguist_languages.sh` is manually copied over.

---

  - Given how `cc_*` targets silently handle C and C++ users likely already expect C support.
  - Add support for formatting C files.

### Test plan

- New test cases added

### DIFF
--- a/docs/format.md
+++ b/docs/format.md
@@ -47,7 +47,7 @@ format_multirun(
 ## languages
 
 <pre>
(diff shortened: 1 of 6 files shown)


alexeagle:
I think the more principled answer is to add the missing language: https://github.com/github-linguist/linguist/blob/559a6426942abcae16b6d6b328147476432bf6cb/lib/linguist/languages.yml#L720-L734

We have other cases where one formatter covers multiple languages.

agoessling:
@alexeagle That does sound better.  Can you point me to an example to pattern match from?

alexeagle:
I think it's simply following the pattern that C++ and other languages here already use.

agoessling:
Ok, I think this is a more principled way of adding C formatting support.

PR #359, opened 2024-08-04
Remove cxxopt flags when linting C sources.

agoessling:
Previously cxxopts were used even if the source was a C file.  Now cxxopts and copts are passed correctly to clang-tidy depending on the file type.

---

### Test plan

Include a C++ only flag in `.bazelrc` e.g. ` build --cxxopt="-std=c++17"` and  lint a C target.


alexeagle (approved):
thanks!

Issue #371, opened 2024-08-21
[Bug]: scalafmt.conf includePaths/excludePaths ignored


vinnybod:
### What happened?

I have a reproduction of the issue here: https://github.com/vinnybod/bazel_java_example/tree/rules_lint_config

TL;DR
I have a scalafmt.conf that specifies which files to format. When running directly via `scala-cli fmt`, it formats the correct files. When running with `bazel run format`, it formats every file.
```
project {
    includePaths = [
        "glob:**/app/**/FormatThis.scala",
    ]
    excludePaths = [
        "glob:**/app/**/DontFormatThis.scala"
    ]
}
```

### Version

Development (host) and target OS/architectures: MacOS

Output of `bazel --version`: 7.2.1

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:

Language(s) and/or frameworks involved:
scala - scalafmt

### How to reproduce
https://github.com/vinnybod/bazel_java_example/tree/rules_lint_config

Run `bazel test //tools/format:format_test --test_output=errors` to see that both `DontFormatThis.scala` and `FormatThis.scala`
 are attempted to be formatted.
 
```shell
==================== Test output for //tools/format:format_test_Scala_with_scalafmt:
--- a/Users/vinnybod/dev/vinnybod/bazel_java_example/app/src/main/scala/com/example/DontFormatThis.scala
+++ b/Users/vinnybod/dev/vinnybod/bazel_java_example/app/src/main/scala/com/example/DontFormatThis.scala
@@ -4,6 +4,4 @@
 
-
-  def main
-(args:
-Array[String]) = println("Hello, world")
+  def main(args: Array[String]) = println("Hello, world")
 }
+
--- a/Users/vinnybod/dev/vinnybod/bazel_java_example/app/src/main/scala/com/example/FormatThis.scala
+++ b/Users/vinnybod/dev/vinnybod/bazel_java_example/app/src/main/scala/com/example/FormatThis.scala
@@ -4,6 +4,4 @@
 
-
-  def main
-(args:
-Array[String]) = println("Hello, world")
+  def main(args: Array[String]) = println("Hello, world")
 }
+
error: --test failed
FAILED: A formatter tool exited with code 1
Try running 'bazel run //tools/format:format_test_Scala_with_scalafmt ' to fix this.
================================================================================
Target //tools/format:format_test_Scala_with_scalafmt up-to-date:
[...]

PR #372, opened 2024-08-21
Add --respect-project-filters to scalafmt

vinnybod:
https://github.com/aspect-build/rules_lint/issues/371
This is a similar solution how rules_lint handles ruff with the `--force-exclude` flag.

---

### Test plan

Follow the reproduction instructions in my example repo before and after this change: https://github.com/vinnybod/bazel_java_example/tree/rules_lint_config
After this change, it only formats the file that is expected to be formatted.


alexeagle:
I reproduced your issue yesterday but didn't make progress - thanks for figuring it out!

alexeagle (approved):

Issue #383, opened 2024-09-07
[Bug]: gitignored files still being formatted


alexeagle:
### What happened?

Files marked in `.gitattributes` like

```
protos/**/*.pb.go	linguist-generated=true
```

are still getting formatted.

### Version

1.0.0

### How to reproduce

I can repro this by making a small change in https://github.com/aspect-build/bazel-examples/blob/main/write_source_files/bar/bar.proto for example. 

`bazel run @@//write_source_files/bar:bar_go_protos.update_go_pb` then updates the `.pb.go` file next to it, and that should be marked as generated thanks to https://github.com/aspect-build/bazel-examples/blob/main/.gitattributes#L8

However when pre-commit runs, it is still reformatting that file, breaking the up-to-date test from the write_source_files rule.

PR #384, opened 2024-09-10
fix(format): Allow git-attributes to be set to true

mrmeku:
Prior to this change, if one of the git attributes used to disable format checks was set to a value
of `true`, it was still formatted since the only
valid syntax was setting the attribute without
any value.

After this change the attributes can be used without a value or with the value of `true`

Fixes https://github.com/aspect-build/rules_lint/issues/383

---

### Test plan

- New test cases added

Git attributes for disabling format checks now support both unset values and `true`. This allows for more flexible configuration when excluding files from formatting.

### DIFF
--- a/format/private/format.sh
+++ b/format/private/format.sh
@@ -162,7 +162,7 @@ function ls-files {
           fi
 
           # Check if the attribute is set
-          if [[ "$line" == *"set" ]]; then
+          if [[ "$line" == *": set" || "$line" == *": true" ]]; then
               attribute_set=true
           fi
       done <<< "$git_attributes"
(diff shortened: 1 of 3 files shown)


format/private/format.sh
```diff
@@ -162,7 +162,7 @@ function ls-files {
           fi
 
           # Check if the attribute is set
-          if [[ "$line" == *"set" ]]; then
+          if [[ "$line" == *": set" || "$line" == *": true" ]]; then
```
alexeagle, line 165: Too bad the syntax sugar didn't get normalized when we ran git

alexeagle (approved):
Thanks!

PR #386, opened 2024-09-11
Use --diff when running ruff as check

TimotheusBachinger:

- when using ruff as formatter and during check, a diff will be displayed in case of files to format


alexeagle (approved):
cool thanks

PR #391, opened 2024-09-16
fix: support file list input for format command to bypass OS limits

### LINKED ISSUE #387: [FR]: formatter: allow arguments spilled into a params file

When there are many files, `pre-commit` or other tooling is going to run `bazel run :format file1 file2 ... fileN` and this exceeds the OS limit on length of the command line.

### Describe the feature

Accept a file that has the arguments. One common idiom for this is an `@` sigil to indicate that a parameter is a flagfile, not an actual file, like
`bazel run :format @these_files.txt`

This does break in the case that users actually want to format a file that starts with an `@` character though.

mrmeku:
Add support for '@file' syntax to read format targets from a file Implement batch processing to handle large numbers of files Maintain compatibility with direct command-line arguments

This change allows the format command to process an arbitrarily large number of files by reading targets from a file, bypassing OS command line length limits. It addresses issues with pre-commit hooks and other tooling when formatting many files at once.
Note: This introduces a limitation where files starting with '@' cannot be directly specified as format targets. Users must use the full file path for such files.

Closes #387

---

### Test plan

Manually ran the script with `@args.txt` which 400,000 lines of `package.json`. The file was formatted 40 times as expected and the OS limit for line length was not reached.

Support for '@file' syntax to read format targets from a file. This allows users to specify more files that would be allowed by the OS line limit (such as during a pre-commit check).

Note: Files starting with '@' now require full path specification

### DIFF
--- a/format/private/format.sh
+++ b/format/private/format.sh
@@ -43,7 +43,7 @@ function on_exit {
       ;;
     *)
       echo >&2 "FAILED: A formatter tool exited with code $code"
-      if [ "$mode" == "check" ]; then
+      if [[ "${mode:-}" == "check" ]]; then
         echo >&2 "Try running '$FIX_CMD' to fix this."
       fi
       ;;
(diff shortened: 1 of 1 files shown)


format/private/format.sh
```diff
@@ -52,6 +52,44 @@ function on_exit {
 
 trap on_exit EXIT
 
+function process_args_in_batches() {
+    local lang="$1"
+    local bin="$2"
+    local flags="$3"
+    shift 3
+    local args=("$@")
+    local max_batch_size=10000
```
alexeagle, line 61: where did you get this number? Add a comment explaining how it's derived.

alexeagle (approved):

PR #392, opened 2024-09-17
chore: point to non-deprecated rustfmt

alexeagle:
See https://github.com/bazelbuild/rules_rust/pull/2703

---

### Test plan


example/src/Bar.java
```diff
@@ -1,8 +1,6 @@
 package src;
 
 // Unused imports are suppressed in suppressions.xml, so this should not raise issue.
-import java.util.Objects;
```
illicitonion, line 4: The comment implies these unused imports are intentional and shouldn't be removed?

illicitonion (approved):

example/src/Bar.java
alexeagle: oh oops, I ran format to test this and should have checked for spurious deltas

Issue #394, opened 2024-09-20
[Bug]: lint_stylelint_aspect does not include config dependencies


sallustfire:
### What happened?

In analogy with eslint, the typical setup for the stylelint config file is to copy it to the output tree as a js_library. The config file may depend on external stylelint plugins which need to be made available via runfiles. The implementation only declares a dependency on the config file and does not include the config file's dependencies.

### Version

Development (host) and target OS/architectures: Pop!_OS 22.04 LTS

Output of `bazel --version`: 7.1.1

Language(s) and/or frameworks involved: stylelint

### How to reproduce

```shell
Add extends to the stylelint config or depend on a plugin.
```

alexeagle:
fixed by #395

PR #395, opened 2024-09-20
fix: Include stylelint config deps (#394)

sallustfire:
If the stylelint config is declared via js_library and has a dependency on other sources or npm packages, these dependencies are now included in the linting action.

---

yes, this fixes a bug in what is a standard pattern

### Test plan

- The existing test case is extended to have a dependency on a npm package in the config


alexeagle (approved):
thank you! nice job adding the missing example usage as regression test

PR #400, opened 2024-10-02
chore: Silence prettier output unless error

honnix:
Silence prettier output unless error because it defaults to `log` that is too verbose: https://prettier.io/docs/en/cli.html#--log-level

---

- Searched for relevant documentation and updated as needed: no update needed
- Breaking change (forces users to change their own code or config): yes because users will no longer see detailed outputs

> `Prettier` is now configured to by default log at `error` level.

### Test plan

### DIFF
--- a/format/private/formatter_binary.bzl
+++ b/format/private/formatter_binary.bzl
@@ -42,7 +42,7 @@ BUILTIN_TOOL_LABELS = {
 CHECK_FLAGS = {
     "buildifier": "-mode=check",
     "swiftformat": "--lint",
-    "prettier": "--check --log-level error",
+    "prettier": "--check --log-level warn",
     "ruff": "format --check --force-exclude --diff",
     "shfmt": "--diff --apply-ignore",
     "java-format": "--set-exit-if-changed --dry-run",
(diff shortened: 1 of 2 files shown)


format/private/formatter_binary.bzl
```diff
@@ -42,7 +42,7 @@ BUILTIN_TOOL_LABELS = {
 CHECK_FLAGS = {
     "buildifier": "-mode=check",
     "swiftformat": "--lint",
-    "prettier": "--check",
+    "prettier": "--check --log-level error",
```
alexeagle, line 45: why error rather than warn?

honnix: Ah, no particular reason, I just happen to have that level patched. I think warn is a better one. Will fix.

alexeagle:
ah, you missed https://github.com/aspect-build/rules_lint/blob/main/example/tools/format/BUILD.bazel#L17-L21

prettier renamed this option at some point, so it belongs in "userland" where you know the flag name

honnix:
> ah, you missed [`main`/example/tools/format/BUILD.bazel#L17-L21](https://github.com/aspect-build/rules_lint/blob/main/example/tools/format/BUILD.bazel?rgh-link-date=2024-10-02T14%3A38%3A57Z#L17-L21)
> 
> prettier renamed this option at some point, so it belongs in "userland" where you know the flag name

Ah, this is nice! We will apply the same then. Thank you!

PR #402, opened 2024-10-02
fix: testing on Windows

alexeagle:
---

### Test plan

- New test cases added


.bazeliskrc
```diff
@@ -1,2 +1,2 @@
-BAZELISK_BASE_URL=https://static.aspect.build/aspect
-USE_BAZEL_VERSION=aspect/2024.39.54
+BAZELISK_BASE_URL=https://github.com/aspect-build/aspect-cli/releases/download
+USE_BAZEL_VERSION=aspect/2024.40.2
```
gregmagolan, line 2: I think you also need to update the `example/.bazeliskrc` to get Windows releases

gregmagolan (approved):

PR #412, opened 2024-10-10
Add missing toolchain dependencies

nradakovic:
clang-tidy needs to be aware of toolchain files in order to find correctly headers. CC toolchain might be configured with sysroot paramter, then sysroot itself needs to be forwarded as input to clang-tidy.

---

### Test plan

- New test cases added


nradakovic:
Sorry wrong target branch

PR #421, opened 2024-10-28
fix: Add implementation_deps to the compilation_context used for clang-tidy

jsharpe:
Updated the compilation_context used by clang-tidy lint actions to include implementation_deps so that any headers that are made available via implementation_deps targets are now visible to clang-tidy.

---

clang-tidy can now see headers provided by `implementation_deps`

### Test plan

Create a C++ target that has an implementation_dep that provides a header file; try to include the header and clang_tidy will fail to find it without this change.


alexeagle (approved):
thanks!

Issue #377, opened 2024-08-30
[FR]: Allow passing arguments to formatters in format_multirun


momilo:
### What is the current behavior?

At the moment it is not possible to provide arguments into the formatters specified in `format_multirun`. Instead, in order to configure their behaviour, one needs to rely on each formatter's internal (default) auto-discovery of configuration files (e.g. .editorconfig for shfmt, .yamlfmt for yamlfmt etc.) or rules_lint's supported  `.gitattributes` annotations (to exclude certain paths - although, in this case, this applies to all formatters, not a specific one).

### Describe the feature

It would be great to able to pass arguments to each individual formatter, to make full use of its features.

This would have a number of benefits:
- generally - not every formatter which supports CLI-based params supports also config files
- avoiding pollution of the repository root with config files - especially for monorepos with multiple languages, this quickly becomes quite an annoyance
- better control over what is included/excluded - some formatters include CLI-based support for what files to include/exclude - this differs between formatters/languages and allows for more precise operation - vs. the currently-available global attribute in `.gitattributes` (applied indiscriminately to all formatters)

alexeagle:
@momilo could you point to some specific motivating examples of cases where a linter takes a CLI flag that you can't put in a config file?

> pollution of the repository root with config files

rules_lint doesn't require this. We only recommend this layout in the example because some tools expect that configurations are in a parent folder of the files being formatted. You're free to put the file wherever you like if the formatter permits.

> control over what is included/excluded

Again, I'd like to see a motivating example of a case where a file should be excluded for one formatter but not another. Wouldn't it be a bad experience if two different formatters are trying to operate on the same file?

Thanks @ewhauser for pointing out an existing workaround: put these in your own wrapper around the binary, then register that one:

```
sh_binary(
     name = "format-sql",
     srcs = [
         "format-sql.sh",
     ],
     data = [
         "//:pyproject.toml",
         "//tools/sql-formatter",
         "@pypi_shandy_sqlfmt//:rules_python_wheel_entry_point_sqlfmt",
     ],
     visibility = ["//visibility:public"],
     deps = [
         "@bazel_tools//tools/bash/runfiles",
     ],
 )
 format_multirun(
     name = "format",
     sql = ":format-sql",
 )
```

momilo:
Hey Alex,
Thank you for the response.

[...]

PR #426, opened 2024-11-19
Add --lint=fix and --lint=warn to buildifier

nlou9:
Add flag --lint=fix and --lint=warn to the buildifier. 
https://github.com/bazelbuild/buildtools/blob/main/WARNINGS.md

---

### Test plan

- New test cases added

### DIFF
--- a/format/private/formatter_binary.bzl
+++ b/format/private/formatter_binary.bzl
@@ -40,7 +40,7 @@ BUILTIN_TOOL_LABELS = {
 
 # Flags to pass each tool's CLI when running in check mode
 CHECK_FLAGS = {
-    "buildifier": "-mode=check",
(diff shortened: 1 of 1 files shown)


alexeagle (commented):
This looks incorrect because Buildifier is run here as a formatter, not a linter. See the README for the design of how they are treated separately

alexeagle:
We don't have a buildifier linter here yet, because the linting is based on an aspect walking the dependency graph, and there are typically no targets that list `BUILD` and `*.bzl` files as their sources. I'm not sure if it's worth introducing `bzl_library` in every package.

alexeagle:
https://github.com/aspect-build/rules_lint/issues/377

alexeagle:
closing as dupe of #377 - thanks for helping out!

PR #432, opened 2024-11-25
fix: Remove implicit requirement on aspect_rules_js

fahhem:
This doesn't remove it from the `MODULE.bazel` since some language aspects do still use `aspect_rules_js`. For the languages that don't rely on it, this allows `bazel` to skip downloading `aspect_rules_js` 

### Test plan


alexeagle (approved):
thanks!

PR #435, opened 2024-11-26
fix: support newer eslint versions

### LINKED ISSUE #434: [Bug]: support for newer ESLint versions
### What happened?

On newer eslint versions strip-ansi and text-table where dropped from it's dependencies but rules_lint depends on it implicitly by requiring them within it's stylish formatter.

### Version

Development (host) and target OS/architectures: linux/arm64

Output of `bazel --version`: bazel 7.4.0

`WORKSPACE` or `MODULE.bazel` file: 1.0.3

Language(s) and/or frameworks involved: eslint

### How to reproduce

```shell
Try to use eslint in version 9.12 or newer.

https://github.com/eslint/eslint/releases/tag/v9.12.0
> refactor: replace strip-ansi with native module

https://github.com/eslint/eslint/releases/tag/v9.15.0
> perf: optimize text-table by replacing regex with trimEnd
```

### Any other information?

I would suggest adding the dependencies into the rules_lint. IMHO it's not a good idea to have a script which has dependency requirements but does not provide them. This would further break as eslint evolves. The formatter should be treated as standalone formatter despite its origin. This would help in further diverging from the original implementation in case this is required.

tobiaskohlbau:
Support more recent eslint versions as up to 9.15.0. Instead of implicilty depending on the dependency tree of eslint itself add dependencies of eslint.stylish-formatter expliclity.

Fixes #434

---

### Test plan

Created repository which has eslint 9.15.0 as requirement. Add local_path_override to the branch of this repository and try to run the linting.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -30,3 +30,10 @@ multitool.hub(lockfile = "//lint:multitool.lock.json")
 use_repo(multitool, "multitool")
 
 bazel_dep(name = "stardoc", version = "0.7.0", dev_dependency = True, repo_name = "io_bazel_stardoc")
+
+npm = use_extension("@aspect_rules_js//npm:extensions.bzl", "npm")
+npm.npm_translate_lock(
+    name = "rules_lint_npm",
+    pnpm_lock = "//lint/js:pnpm-lock.yaml",
+)
+use_repo(npm, "rules_lint_npm")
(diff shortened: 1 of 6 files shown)


tobiaskohlbau:
This is just my idea of solving #435 but there exist more options.

alexeagle (commented):
Thanks - but it's quite undesirable for us to publish a Bazel module that also has its own npm dependencies. It gets very confusing for users when our versions don't match theirs, don't appear in their lockfile, don't come from their artifactory, etc.

We should instead move our stylish-formatter to its own package, or have a different vendored version of it for different eslint versions.

tobiaskohlbau:
Totally reasonable to not have internals within the global rules_lint package. Let's drop this PR in favor of your solution! Thanks for the quick heads up!

PR #441, opened 2024-12-02
perf: avoid use of depset.to_list

jbedard:
---

### Test plan


alexeagle:
Awesome, thanks - is this related to our customer pointing out the slowness?

Failure:
```
#   	File "/home/runner/.cache/bazel/_bazel_runner/89e91fd6b6dccc7b506d63833e3c5c70/external/aspect_rules_lint~/lint/stylelint.bzl", line 147, column 44, in stylelint_fix
#   		inputs = _gather_inputs(ctx, srcs) + [patch_cfg],
#   Error: unsupported binary operation: depset + list
``` 
LMK if you'd like any help greening up

alexeagle (approved):

jbedard:
It's not urgent so I won't do a release atm. If we're pinged about releasing we can do it then.

Issue #406, opened 2024-10-08
[Bug]: clang-tidy aspect fails on CcInfo rules without srcs


cramertj:
### What happened?

https://github.com/aspect-build/rules_lint/blob/559ed12caf62bb4b6a5048b495b86e611fb7a359/lint/clang_tidy.bzl#L309-L312

assumes that all `CcInfo` rules have a `srcs` label list:

https://github.com/aspect-build/rules_lint/blob/559ed12caf62bb4b6a5048b495b86e611fb7a359/lint/clang_tidy.bzl#L145-L149

In my project (Pigweed) we have custom `pwpb_proto_library` rule which provides `CcInfo` but does *not* have a `srcs` list:

https://github.com/google/pigweed/blob/8bd77aba07ab3dce5220b23994cd3ecfbcefda10/pw_protobuf_compiler/pw_proto_library.bzl#L536-L549

This leads to the following error when running `bazelisk lint //pw_rpc:echo_cc.pwpb`:

```
➜  pigweed git:(af6a29e26) ✗ bazelisk lint //pw_rpc:echo_cc.pwpb
2024/10/08 14:55:38 Using unreleased version at commit 1ff5af18d045e8f30a2a0367470db4e8225a785c
2024/10/08 14:55:38 Using unreleased version at commit 1ff5af18d045e8f30a2a0367470db4e8225a785c
ERROR: /usr/local/google/home/cramertj/src/pigweed/pw_rpc/BUILD.bazel:486:17: in //tools/lint:linters.bzl%clang_tidy aspect on _pwpb_proto_library rule //pw_rpc:echo_cc.pwpb:
Traceback (most recent call last):
	File "/tmp/bazel_output_user_root/02fd49830a521506a99f19a1e23f8759/external/aspect_rules_lint+/lint/clang_tidy.bzl", line 312, column 33, in _clang_tidy_aspect_impl
		files_to_lint = _filter_srcs(ctx.rule)
	File "/tmp/bazel_output_user_root/02fd49830a521506a99f19a1e23f8759/external/aspect_rules_lint+/lint/clang_tidy.bzl", line 149, column 38, in _filter_srcs
		return [s for s in rule.files.srcs if _is_source(s)]
Error: No attribute 'srcs' in files. Make sure there is a label or label_list type attribute with this name
[...]

PR #447, opened 2024-12-09
fix: Add option to skip target without srcs

johan-vcc:
Description
---
Add option 'skip_if_no_srcs' to lint_clang_tidy aspect.
Set to True to skip target if it provides CcInfo but has no source files. By default this will generate an error.

See issue #406 

Adding an option to skip these targets enables us to run `./lint.sh //package//...` even if some target has no source files, e.g. code generation targets.

Add option 'skip_if_no_srcs' to lint_clang_tidy aspect.

### Test plan

bazel build with lint_clang_tidy_aspect for a target that provides 'CcInfo' but does not have 'srcs' attribute.
With skip_if_no_srcs set to True:
Build completes with empty `target.AspectRulesLintClangTidy.out` file.
Without new option / new option set to False:
```shell
File ".../aspect_rules_lint~/lint/clang_tidy.bzl", line 186, column 38, in _filter_srcs
                return [s for s in rule.files.srcs if _is_source(s)]
Error: No attribute 'srcs' in files. Make sure there is a label or label_list type attribute with this name
```

### DIFF
--- a/lint/clang_tidy.bzl
+++ b/lint/clang_tidy.bzl
@@ -177,10 +177,12 @@ def _is_source(file):
     return (file.is_source and file.extension in permitted_source_types)
 
 # modification of filter_srcs in lint_aspect.bzl that filters out header files
-def _filter_srcs(rule):
+def _filter_srcs(rule, skip_if_no_srcs = False):
     if "lint-genfiles" in rule.attr.tags:
         return rule.files.srcs
     else:
+        if skip_if_no_srcs and not hasattr(rule.files, "srcs"):
+            return []
         return [s for s in rule.files.srcs if _is_source(s)]
 
 def is_parent_in_list(dir, list):
@@ -369,7 +371,7 @@ def _clang_tidy_aspect_impl(target, ctx):
     if not CcInfo in target:
         return []
 
-    files_to_lint = _filter_srcs(ctx.rule)
+    files_to_lint = _filter_srcs(ctx.rule, ctx.attr._skip_if_no_srcs)
     compilation_context = target[CcInfo].compilation_context
     if hasattr(ctx.rule.attr, "implementation_deps"):
         compilation_context = cc_common.merge_compilation_contexts(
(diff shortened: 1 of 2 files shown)


alexeagle:
Thanks - is there ever a reason users ought to set `skip_if_no_srcs=False` ? If not, can we make a simpler change to always exclude rather than introduce a new configuration option?

johan-vcc:
I am unsure of other use-cases. That is why I added the option in case the old behavior is needed.
If you think a simpler always-exclude is good, that would work for my use-case.

alexeagle:
@jsharpe can you think of any reason the aspect might need to visit a library with empty `srcs`? Or is there a better attribute on `CcInfo` that tells us it's a no-op?

jsharpe:
I don't think it would need to visit the target if it has an empty `srcs` since this is where we construct the list of files to apply clang-tidy to anyway so if that was empty there is nothing to do. IMO we can always skip if `srcs` is empty.

lint/clang_tidy.bzl
```diff
@@ -369,7 +371,7 @@ def _clang_tidy_aspect_impl(target, ctx):
     if not CcInfo in target:
         return []
 
-    files_to_lint = _filter_srcs(ctx.rule)
+    files_to_lint = _filter_srcs(ctx.rule, ctx.attr._skip_if_no_srcs)
     compilation_context = target[CcInfo].compilation_context
```
jsharpe, line 375: Now I look at it - compilation_context could be None - see https://bazel.build/rules/lib/providers/CcInfo#CcInfo; I suspect that this will happen precisely when there are no srcs present and no deps? And if so we might need to protect other places further down where we access compilation_context?

alexeagle: @johan-vcc does it fix the issue for you if you check for this being `None` ?

johan-vcc: Tried it out, then that check would need to be before it tries to access rule.files.srcs in _filter_srcs().
However, in my case the targets have compilation_context so that would not avoid the need to check srcs as well.

PR #448, opened 2024-12-11
fix: Use relative paths with stylelint

JesseTatasciore:
Patch the stylelint compact formatter so that it outputs relative paths. This is needed as by default we were getting paths that included the bazel-out directory

---

Stylelint will now output relative, non-bazel paths

### Test plan

I ran this using the aspect CLI and produced the following result:
```
Patch for //src:css from linter mnemonic AspectRulesLintStylelint
Lint results for //src:css:

src/hello.css: line 11, col 5, error - Unexpected empty block (block-no-empty)
src/hello.css: line 16, col 3, error - Unexpected duplicate "color" (declaration-block-no-duplicate-properties)

Some problems have automated fixes available:

  src/hello.css | 1 -
  1 file, 1 deletion(-)
```

### DIFF
--- a/lint/BUILD.bazel
+++ b/lint/BUILD.bazel
@@ -128,6 +128,12 @@ alias(
     visibility = ["//visibility:public"],
 )
 
+alias(
+    name = "stylelint.compact-formatter",
+    actual = "//lint/js:stylelint.compact-formatter",
+    visibility = ["//visibility:public"],
+)
+
 bzl_library(
     name = "eslint",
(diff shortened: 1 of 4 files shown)


alexeagle (approved):
Could you link to the issue on stylelint that it doesn't support relative paths (file one if it doesn't exist?)

JesseTatasciore:
Stylelint issue: https://github.com/stylelint/stylelint/issues/8204

PR #463, opened 2025-01-21
feat: Add keep-sorted as a linter

fzakaria:
Add Google's [keep-sorted](https://github.com/google/keep-sorted) as linter.

**Motivation**
There are plenty of times when you want a list sorted it's not part of the language formatted.
> keep-sorted is a language-agnostic formatter that sorts lines between two markers in a larger file.

I have added a failure case in `Bar.java` that you can see if you run it.
```shell
> bazel build --aspects=//tools/lint:linters.bzl%keep_sorted //src:bar --remote_download_regex='.*AspectRulesLint.*' --output_groups=rules_lint_human --@aspect_rules_lint//lint:fail_on_violation
INFO: Invocation ID: 0c3aa47a-092c-4504-a999-acdccf297733
INFO: Analyzed target //src:bar (1 packages loaded, 11 targets configured).
ERROR: /Users/fzakaria/code/github.com/aspect-build/rules_lint/example/src/BUILD.bazel:74:13: Linting //src:bar with KeepSorted failed: (Exit 1): bash failed: error executing AspectRulesLintKeepSorted command (from target //src:bar) /bin/bash -c ... (remaining 5 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
[
  {
    "path": "src/Bar.java",
    "lines": {
      "start": 9,
      "end": 12
    },
    "message": "These lines are out of order.",
    "fixes": [
      {
        "replacements": [
          {
            "lines": {
              "start": 9,
              "end": 12
            },
            "new_content": "    A(),\n    B(),\n    C(),\n    D(),\n"
          }
        ]
      }
    ]
  }
]
Aspect //tools/lint:linters.bzl%keep_sorted of //src:bar failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 0.453s, Critical Path: 0.14s
INFO: 2 processes: 2 internal.
ERROR: Build did NOT complete successfully
```

Changes:
* Added go.mod and gazelle to example directory
* Added keep_sorted aspect and relevant files
* Updated stardoc generated markdown files


alexeagle:
Checked that this builds from source as a workaround for https://github.com/google/keep-sorted/issues/37

README.md
```diff
@@ -47,6 +47,7 @@ New tools are being added frequently, so check this page again!
 | TSX                    | [Prettier]            | [ESLint]                         |
 | TypeScript             | [Prettier]            | [ESLint]                         |
 | YAML                   | [yamlfmt]             |                                  |
+| Everything             |                       | [keep-sorted]                    |
```
hofbi, line 50: Isn't this more a formatter than a linter?

fzakaria: There's no way I could see to add an Everything formatter given the current setup; since it's 1 tool to 1 language
keep-sorted has a fix and a lint mode; so it definitely works with lint as well. 🙃 
It felt like the most straightforward way to add it to the rule albeit it would be nice to add it as a formatter too.

happy to take suggestions.

alexeagle: changing the order of lines can have side-effects, so it's not a formatter. As the top of the README explains, formatters should be blindly applyable without alternatives or testing.

example/main.go
```diff
@@ -0,0 +1,7 @@
+// The tools module is used to keep tool dependencies separate from the main dependencies of the repo
+// For more details see: https://github.com/golang/go/issues/25922#issuecomment-1038394599
```
alexeagle, line 2: I'll change this to follow https://github.com/bazel-contrib/rules_go/blob/7f6a9bf5870f2b5ffbba1615658676dcabf9edc7/docs/go/core/bzlmod.md#depending-on-tools instead

That's a better pattern for users to follow, so that keep-sorted dependencies don't get constraint-solved along with their application dependencies.

Also we can provide this go.mod file from rules_lint, so as long as users are okay with the version of keep-sorted we have, they can skip some steps.

example/tools/lint/linters.bzl
```diff
@@ -118,3 +119,9 @@ spotbugs = lint_spotbugs_aspect(
 )
 
 spotbugs_test = lint_test(aspect = spotbugs)
+
+keep_sorted = lint_keep_sorted_aspect(
+    binary = "@@gazelle~~go_deps~com_github_google_keep_sorted//:keep-sorted",
```
alexeagle, line 124: I'll fix this up - you shouldn't rely on the canonical name (for example the `~` changed to a `+`

should always use the apparent name and let Bazel resolve it.

alexeagle (commented):
Thanks! I'll pick this up now, and add support for the auto-fixer.

example/.bazelrc
```diff
@@ -1,6 +1,9 @@
 # Automatically apply --config=linux, --config=windows etc
 common --enable_platform_specific_config
 
+# Permit MODULE.bazel to use_extension("@gazelle//:extensions.bzl", "go_deps", isolate = True)
```
fzakaria, line 4: What's this?

example/tools/lint/linters.bzl
```diff
@@ -113,8 +114,14 @@ clang_tidy_global_config = lint_clang_tidy_aspect(
...
-    binary = "@@//tools/lint:spotbugs",
-    exclude_filter = "@@//:spotbugs-exclude.xml",
+    binary = Label("@//tools/lint:spotbugs"),
+    exclude_filter = Label("@//:spotbugs-exclude.xml"),
 )
 
 spotbugs_test = lint_test(aspect = spotbugs)
+
+keep_sorted = lint_keep_sorted_aspect(
+    binary = Label("@com_github_google_keep_sorted//:keep-sorted"),
```
fzakaria, line 124: Was i missing `Lablel()` so it would make apparent -> canonical?
I thought I could just write the apparent as a string and it would change it as necessary.

fzakaria:
Thanks for taking it over the line.

example/tools/lint/linters.bzl
alexeagle: The problem is which context strings are evaluated to labels.
When you pass a string to a macro or function, it is evaluated in the receiving context, in this case that's in rules_lint which won't have this apparent repository.
The explicit Label() call here ensures it's evaluated in this repository context.

Issue #454, opened 2024-12-23
[FR]: Improve checkstyle to support different checkstyle.xml configs in a repo


nlou9:
### What is the current behavior?

Checkstyle does not automatically take the nearest configuration file for different modules. We need to explicitly specify the configuration file for each module in your build tool configuration.

### Describe the feature

checkstyle could handle different configuration files for different modules in one repo.

PR #464, opened 2025-01-21
Support multiple configuration files for different modules in checkstyle

nlou9:
In some use case, different modules need different checkstyle configuration files as described in this feature request. 
https://github.com/aspect-build/rules_lint/issues/454

This PR :
- upgrade the bazel version to 7.4.1 which would support data type `string_keyed_label_dict`
- add `configs` attribute which is a map of module label and configuration file. 

Test:
In the test, add one checkstyle_subdir.xml file for src/subdir which only differs with different line length. 
```
bazel lint //src/subdir:bar
INFO: Analyzed target //src/subdir:bar (0 packages loaded, 0 targets configured).
INFO: From Linting //src/subdir:bar with Checkstyle:
Checkstyle ends with 1 errors.
INFO: Found 1 target...
Aspect //tools/lint:linters.bzl%checkstyle of //src/subdir:bar up-to-date:
  bazel-bin/src/subdir/bar.AspectRulesLintCheckstyle.out
  bazel-bin/src/subdir/bar.AspectRulesLintCheckstyle.out.exit_code
INFO: Elapsed time: 0.380s, Critical Path: 0.21s
INFO: 2 processes: 1 internal, 1 darwin-sandbox.
INFO: Build completed successfully, 2 total actions
Lint results for //src/subdir:bar:

Starting audit...
[ERROR] /private/var/tmp/_bazel_louna/3cc840f4a4f44129661666d5b35a9196/sandbox/darwin-sandbox/14/execroot/_main/src/subdir/Bar.java:3: Line is longer than 84 characters (found 85). [LineLength]
Audit done.
```
```
bazel lint //src:bar
INFO: Analyzed target //src:bar (0 packages loaded, 2 targets configured).
INFO: Found 1 target...
Aspect //tools/lint:linters.bzl%checkstyle of //src:bar up-to-date:
  bazel-bin/src/bar.AspectRulesLintCheckstyle.out
  bazel-bin/src/bar.AspectRulesLintCheckstyle.out.exit_code
INFO: Elapsed time: 0.098s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
Lint results for //src:bar:

Starting audit...
[ERROR] /private/var/tmp/_bazel_louna/3cc840f4a4f44129661666d5b35a9196/sandbox/darwin-sandbox/13/execroot/_main/src/Bar.java:3: Line is longer than 20 characters (found 85). [LineLength]
[ERROR] /private/var/tmp/_bazel_louna/3cc840f4a4f44129661666d5b35a9196/sandbox/darwin-sandbox/13/execroot/_main/src/Bar.java:8: Line is longer than 20 characters (found 59). [LineLength]
[ERROR] /private/var/tmp/_bazel_louna/3cc840f4a4f44129661666d5b35a9196/sandbox/darwin-sandbox/13/execroot/_main/src/Bar.java:9: Line is longer than 20 characters (found 35). [LineLength]
Audit done.
```

---

### Test plan

- New test cases added

### DIFF
--- a/docs/checkstyle.md
+++ b/docs/checkstyle.md
@@ -110,3 +110,23 @@ Attrs:
 | <a id="lint_checkstyle_aspect-rule_kinds"></a>rule_kinds |  <p align="center"> - </p>   |  `["java_binary", "java_library"]` |
 
 
+<a id="rebuild_map"></a>
+
+## rebuild_map
+
+<pre>
+load("@aspect_rules_lint//lint:checkstyle.bzl", "rebuild_map")
+
+rebuild_map(<a href="#rebuild_map-configs">configs</a>)
+</pre>
+
+
+
+**PARAMETERS**
+
+
+| Name  | Description | Default Value |
+| :------------- | :------------- | :------------- |
+| <a id="rebuild_map-configs"></a>configs |  <p align="center"> - </p>   |  none |
+
+


lint/checkstyle.bzl
```diff
@@ -82,6 +83,15 @@ def _checkstyle_aspect_impl(target, ctx):
...
     files_to_lint = filter_srcs(ctx.rule)
+    config_file = ctx.file._config
+    if ctx.attr._configs:
+        label = str(target.label)
+        keys = sorted(ctx.attr._configs.keys(), key=len, reverse = True)
+        for key in keys:
+            if label.startswith(key):
+                config = ctx.attr._configs[key]
+                config_file = config.files.to_list()[0]
+                break
```
fzakaria, line 94: do you want longest matching prefix?

alexeagle: I'd expect it to use the config from the nearest ancestor, matching most other tools

.bazelversion
```diff
@@ -1,4 +1,4 @@
-7.1.1
+7.4.1
```
alexeagle, line 1: this is problematic, will users be forced to upgrade Bazel as well? In that case it's a breaking change and hard to rollout. We should look at alternatives that work with older Bazel, and/or whether we can have some conditional logic using bazel_features so the new way is only enabled when the Bazel version supports it

alexeagle (commented):
I think this is a workaround for a flaw in checkstyle, which I would expect to be able to resolve the right configuration for a given source file on its own. We should link to an issue there (creating it if needed) so it's clear that we're working around something, and in the event they fix it upstream we'll be able to remove our workaround.

lint/checkstyle.bzl
alexeagle: How about a prefactoring where you move your checkstyle configs closer to the folder they govern? Having them all in a single directory seems like it makes it hard to reason about, even without taking Bazel into account. If someone wanted to update the checkstyle config for their code, how would they figure out which one to edit?

example/tools/lint/linters.bzl
```diff
@@ -52,9 +52,13 @@ pmd = lint_pmd_aspect(
 pmd_test = lint_test(aspect = pmd)
 
 checkstyle = lint_checkstyle_aspect(
-    binary = Label("@//tools/lint:checkstyle"),
-    config = Label("@//:checkstyle.xml"),
-    data = [Label("@//:checkstyle-suppressions.xml")],
+    binary = "@@//tools/lint:checkstyle",
+    config = "@@//:checkstyle.xml",
+    configs = {
+        "@@//:checkstyle.xml": "@@//src,@@//test",
```
alexeagle, line 58: I'm taking another look at what apple_rules_lint did to permit users to configure linters per-directory, i.e. https://github.com/apple/apple_rules_lint/blob/main/api.md#get_lint_config allows you to set `tags` and https://github.com/apple/apple_rules_lint/blob/main/lint/private/package_lint_config.bzl#L14 allows the linter configuration to be selected in the same package being visited.

I'm curious, you must have considered just using apple_rules_lint instead of Aspect, since this checkstyle case is the example on their main README and I assume it motivated their design for selecting configs. Did you come to understand any tradeoffs between these approaches?

PR #474, opened 2025-02-02
feat: add XML format via prettier

b0inbot:
Added XML as a supported formatter and make it use prettier by default. I am not interested in using prettier myself but it's already configured so I might as well. 

### Test plan

- New test cases added
- Tested on my own repo with a custom xml formatter.  
   - Commit is here but i do not expect you to run it yourself https://github.com/b0inbot/tools/commit/0b13ed07555c8e8081055cd41dd30c8cc9bcb3d6
- unable to test via bazel run format within the example/ folder (Error running code complaining about libcrypt so.) Hoping CI/CD within ubuntu can run the test case.
- test folder does work:

```
$ pwd
/home/boin/Projects/tools/tools/rules_lint/format/test
$ bazel run format 2>&1 | grep -i xml
+ prettier --write example/checkstyle-suppressions.xml example/checkstyle.xml example/ktlint-baseline.xml example/pmd.xml example/spotbugs-exclude.xml example/src/hello.xml
Formatted XML in 0m0.011s
```

### DIFF
--- a/format/private/format.sh
+++ b/format/private/format.sh
@@ -142,6 +142,7 @@ function ls-files {
       'TypeScript') patterns=('*.ts' '*.cts' '*.mts') ;;
       'Vue') patterns=('*.vue') ;;
       'YAML') patterns=('*.yml' '*.yaml' '.clang-format' '.clang-tidy' '.gemrc') ;;
+      'XML') patterns=('*.xml') ;;
 
       # Note: terraform fmt cannot handle all HCL files such as .terraform.lock
       # "Only .tf and .tfvars files can be processed with terraform fmt"
--- a/format/private/format.sh
+++ b/format/private/format.sh
@@ -142,7 +142,9 @@ function ls-files {
       'TypeScript') patterns=('*.ts' '*.cts' '*.mts') ;;
       'Vue') patterns=('*.vue') ;;
       'YAML') patterns=('*.yml' '*.yaml' '.clang-format' '.clang-tidy' '.gemrc') ;;
-      'XML') patterns=('*.xml') ;;
+      # Note: https://github.com/github-linguist/linguist/blob/559a6426942abcae16b6d6b328147476432bf6cb/lib/linguist/languages.yml#L7767-L7882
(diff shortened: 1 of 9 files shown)


format/private/format.sh
```diff
@@ -142,6 +142,7 @@ function ls-files {
       'TypeScript') patterns=('*.ts' '*.cts' '*.mts') ;;
       'Vue') patterns=('*.vue') ;;
       'YAML') patterns=('*.yml' '*.yaml' '.clang-format' '.clang-tidy' '.gemrc') ;;
+      'XML') patterns=('*.xml') ;;
```
alexeagle, line 145: hm, so far I've been mirroring what GitHub considers to be the filename patterns for a given language. Not surprising that XML is a giant list though https://github.com/github-linguist/linguist/blob/559a6426942abcae16b6d6b328147476432bf6cb/lib/linguist/languages.yml#L7767-L7882

Somehow they also permit duplicates, like `".tsx"` is in this list.

I recognize a couple of these as things we should include here (like ".xsd") but it will take a little extra thought, and it's sad to have to make such a decision here instead of blindly following a "trusted" external source

b0inbot: Oh that's a shame. 

Are patterns configerable at all or is that a feature you may want for this project? might be worth sticking with a baseline: xml, xsd, xslt, xsl.

b0inbot: re configerable patterns: here is what that might look like. 

https://github.com/aspect-build/rules_lint/compare/main...b0inbot:rules_lint:poc/extensions_json

alexeagle: That looks like a great approach to me. I had been thinking of encoding the extensions in starlark but JSON is a lot simpler. I created #484 with your commit and I'll do a bit more on top of it. Thanks!

alexeagle (approved):

alexeagle:
Ah, I found this didn't actually function because the new plugin wasn't declared as a dependency. Will fix in a followup

PR #477, opened 2025-02-07
fix: clang_tidy - Add defines and local_defines attributes to commandline args

molar:
The clang_tidy aspect did not read the `defines` and `local_defines` attribute of the rule, causing clang_tidy to error 
on parsing code that references these defines. This PR extends the clang_tidy.bzl with support for reading these attributes and adds them to the clang-tidy commandline. 

I also added a guard against rules that do not have a `srcs` attribute, but do return a CcInfo. 
---

### Test plan

Extended the example with a `local_defines` attribute and tried to lint the code without my changes, where i got the 
error 
```
_main/src/cpp/lib/hello-time.cc:16:28: error: use of undeclared identifier 'LOCAL_DEFINE_IS_DEFINED' [clang-diagnostic-error]
  static const char* str = LOCAL_DEFINE_IS_DEFINED;
```

after my change the error is gone and the normal fixes are proposed when running `bazel lint //src/cpp/...`.

I only tested on ubuntu x86.

### DIFF
--- a/lint/clang_tidy.bzl
+++ b/lint/clang_tidy.bzl
@@ -273,10 +273,10 @@ def _get_compiler_args(ctx, compilation_context, srcs):
         args.append("-D" + define)
     for define in compilation_context.local_defines.to_list():
         args.append("-D" + define)
-    if hasattr(ctx.rule.attr,"defines"):
+    if hasattr(ctx.rule.attr, "defines"):
         for define in ctx.rule.attr.defines:
             args.append("-D" + define)
-    if hasattr(ctx.rule.attr,"local_defines"):
+    if hasattr(ctx.rule.attr, "local_defines"):
         for define in ctx.rule.attr.local_defines:
             args.append("-D" + define)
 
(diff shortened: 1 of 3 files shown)


alexeagle (approved):
thanks!

lint/clang_tidy.bzl
```diff
@@ -270,6 +273,12 @@ def _get_compiler_args(ctx, compilation_context, srcs):
         args.append("-D" + define)
     for define in compilation_context.local_defines.to_list():
         args.append("-D" + define)
+    if hasattr(ctx.rule.attr, "defines"):
```
alexeagle, line 276: hmm, is it a bug that the compilation_context didn't already supply this local_define value? @jsharpe in case you have an opinion

alexeagle:
@molar are you able to click through the Contributor License Agreement?

molar:
> @molar are you able to click through the Contributor License Agreement?

I am waiting for legal to look it over, not sure of the ETA.

lint/clang_tidy.bzl
jsharpe: Certainly seems like it, but given its the behaviour of bazel releases then we've got to fix up for it; there shouldn't be too much harm (other than line length and the associated increase in memory use) in adding the defines to the commandline multiple times?

molar:
> > @molar are you able to click through the Contributor License Agreement?
> 
> I am waiting for legal to look it over, not sure of the ETA.

@alexeagle  i have signed cla - is there any other issues you would like to be addressed

PR #491, opened 2025-02-28
Add Gherkin Support

ericmcbride:
Added Gherkin (Feature File) support.  We currently use `prettier-plugin-gherkin` for linting our feature files, and we use the aspect_rules_lint for everything else.  Wanted to add this feature so we don't have to maintain a one off.

Adds a new feature for users to use `prettier-plugin-gherkin` w `prettier` to add functionality to formatting feature files.

### Test plan

- Added an example

### DIFF
--- a/example/src/Bar.java
+++ b/example/src/Bar.java
@@ -1,6 +1,8 @@
 package src;
 
 // Unused imports are suppressed in suppressions.xml, so this should not raise issue.
+import java.util.Objects;
+import java.io.BufferedInputStream;
 
 public class Bar {
 
--- a/example/src/Bar.java
+++ b/example/src/Bar.java
@@ -1,8 +1,6 @@
 package src;
 
 // Unused imports are suppressed in suppressions.xml, so this should not raise issue.
-import java.util.Objects;
-import java.io.BufferedInputStream;
 
 public class Bar {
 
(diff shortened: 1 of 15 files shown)


example/src/Bar.java
```diff
@@ -1,8 +1,6 @@
 package src;
 
 // Unused imports are suppressed in suppressions.xml, so this should not raise issue.
-import java.util.Objects;
```
ericmcbride, line 4: These were removed when running `bazel run //:format` so im not sure if the supressions.xml are working correctly.

ericmcbride:
@alexeagle

format/private/format.sh
```diff
@@ -145,7 +145,7 @@ function ls-files {
       # Note: https://github.com/github-linguist/linguist/blob/559a6426942abcae16b6d6b328147476432bf6cb/lib/linguist/languages.yml#L7767-L7882
       # has a giant list of patterns. We arbitrarily choose some "common" ones.
       'XML') patterns=('*.xml' '*.xsd') ;;
-
+      'Gherkin') patterns=('*.feature') ;;
```
alexeagle, line 148: could you include `.story` as well per https://github.com/github-linguist/linguist/blob/559a6426942abcae16b6d6b328147476432bf6cb/lib/linguist/languages.yml#L2371-L2381 ?

example/src/Bar.java
alexeagle: I think something about the Java formatter is misbehaving and making non-format changes. Could you just revert the .java changes pls?

alexeagle (commented):
Thanks!

example/src/Bar.java
ericmcbride: On it

ericmcbride:
@alexeagle I made the changes.  Im running the 2 tests that failed locally and im running the suggestions but they are not working.  I didnt touch the files so im confused whats going on there

alexeagle (approved):

PR #494, opened 2025-03-07
Fix spotbugs warnings for missing classes

vinnybod:
We noticed some non-fatal warnings when running spotbugs about missing classes.
This uses -auxclasspath to add the compile time jars to remove those warnings.

Example:
```
The following classes needed for analysis were missing:
  org.apache.kafka.common.Uuid
```

We also noticed this same issue happens in rules_jvm's spotbugs rule as well.

Note: This fixes the errors for java_library but not java_binary because the JavaInfo provider lacks compile time jar information for binaries.

### Test plan

- New test cases added


alexeagle (approved):
Thanks! Good cleanups too

PR #498, opened 2025-03-14
feat: configure new publish-to-bcr workflow

alexeagle:
Practice run to confirm new end-user documentation.

---

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,32 @@
+name: Publish To BCR
+
+on:
+  # Run the publish workflow after a successful release
+  # Can be triggered from the release.yaml workflow
+  workflow_call:
+    inputs:
+      tag_name:
+        required: true
+        type: string
+  # In case of problems, let release engineers retry by manually dispatching
+  # the workflow from the GitHub UI
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        required: true
+        type: string
+
+permissions:
+  id-token: write
+  attestations: write
+  contents: write
+
+jobs:
+  publish:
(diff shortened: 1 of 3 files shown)


kormide:
The publish workflow will probably fail unless the slsa-verifier issues have been worked out. But if the Publish to BCR app is still installed, then a second working entry should also be opened on release.

kormide (approved):

alexeagle:
Yeah I'm fine with failures here as long as I can reason we've gotten further than the documentation requires

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+jobs:
+  publish:
+    uses: bazel-contrib/publish-to-bcr/.github/workflows/publish.yaml@165f8b872b32a54bcc3c77156db4a74f54a51986 # 2025 March 14
+    with:
+      tag_name: ${{ inputs.tag_name }}
+      # GitHub repository which is a fork of the upstream where the Pull Request will be opened.
+      registry_fork: aspect-build/bazel-central-registry
+    secrets:
+      publish_token: ${{ secrets.PUBLISH_TOKEN }}
```
kormide, line 32: Double checking that you have added this repository secret.

kormide (changes requested):
The new release workflow needs the same permissions as the publish one, because it attests the release artifacts.

.github/workflows/publish.yaml
alexeagle: I'm first going to verify the DX when it's absent so I follow along the user journey "sad paths" too. Want to see if the error message is good or needs any docs

alexeagle:
Actually `attestations: write` is only on release, not publish - the latter just moves the files around

kormide (approved):

alexeagle:
Note, once this new file is on `main` it will appear in the UI and i'll be able to test subsequent fixes from a branch

PR #506, opened 2025-03-25
typo fixed in spotbugs.bzl

fzakaria:


alexeagle:
I did this in #505 as well :)

PR #507, opened 2025-03-26
fix(format): Handle successfuly folders name which contain spaces

ramzesucr:
Fixing format which doesn't handle properly folders names with spaces. When format is executed it fails since the folder is not found because it cuts name with spaces thus it's not complete folder name.

Example of folder: `complex_folder/{{ cookiecutter.name }}/src.go`

---

### Test plan

- New test cases added


alexeagle (approved):
Thanks!

PR #508, opened 2025-03-26
chore: fix release

alexeagle:
Appu found that the slsa-verifier hard-codes the usual semver pattern vX.Y.Z. I was too cute using a shorter one that looks like a GitHub actions major/minor.

Also fix permissions and secret.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -9,6 +9,9 @@ on:
       tag_name:
         required: true
         type: string
+    secrets:
+      publish_token:
+        required: true
   # In case of problems, let release engineers retry by manually dispatching
   # the workflow from the GitHub UI
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yaml
```diff
@@ -9,6 +9,9 @@ on:
       tag_name:
         required: true
         type: string
+    secrets:
+      publish_token:
```
kormide, line 13: Does this need to match the case of the secret, which is `PUBLISH_TOKEN`?

kormide (approved):

.github/workflows/publish.yaml
alexeagle: apparently not, I did test it by triggering a release workflow at this commit: https://github.com/aspect-build/rules_lint/actions/runs/14095611671

PR #512, opened 2025-04-03
clang-tidy emit action per file to lint and merge stdout and error codes

molar:
Some cc_library/binary targets may contain 10+ files listed in sources and as clang-tidy runs single threaded, this can lead 
to long running linting actions. This PR emits a clang-tidy action per source file, so they can be parallelised locally/remotely. The stdout/stderror is merged in unspecified order according to the order of the srcs using a simple 
concatenation. The exit_code of the individual action, is merged by sorting them numerically and selecting the highest number. 

I did also consider making the splitting optional, in case we are unsure of the impact of this change. Let me know if i should do that. 

I also considered updated the progress message to use the source file short path instead, but decided to leave it out here. Currently the progress message will be the same for each individual action and makes it harder to see which source file is being linted.

### Test plan

- Manual testing was also done, both in our own bazel repo and in the examples folder. It has only been tested on linux, so there could be issues with the behaviour of `sort` and `head` on macOS.

### DIFF
--- a/example/src/cpp/main/BUILD
+++ b/example/src/cpp/main/BUILD
@@ -1,5 +1,5 @@
 load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library")
-
+load("//tools/lint:linters.bzl", "clang_tidy_test")
 cc_library(
     name = "hello-greet",
     srcs = ["hello-greet.cc"],
@@ -15,6 +15,11 @@ cc_binary(
         ":hello-greet",
         "//src/cpp/lib:hello-time",
     ],
+)
+
+clang_tidy_test(
+    name = "hello-world-clang-tidy-test",
+    srcs = ["hello-world"],
 )
 
 cc_library(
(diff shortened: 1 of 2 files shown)


molar:
Friendly ping @alexeagle  do you have some input or can you kindly suggest a reviewer?

alexeagle:
I'm sorry - I looked through this once and then got in my head that it was blocked on something.

It's a bit complex/gross to have to merge the exit codes back together this way. like you say, it's using non-hermetic system tools, plus space delimiting is risky when filenames can have spaces, so may need some quoting work.

Is it possible to move the splitting earlier? I'm imagining the aspect visits a cc_library target and produces an action per file at that point. Bazel sees all the actions and can react to the non-zero exit code(s) itself. I think there's no need to aggregate the results back to the library level, right?

molar:
@alexeagle  Thank you for your time on this. I will not be working on this for the time being unfortunately.

PR #514, opened 2025-04-04
feat: Introduce rust diagnostic formatter

mrmeku:
This formatter will be used by a new rust aspect such that when rules
rust outputs clippy's JSON diagnostics, they can be supported by
rules_lint

---

### Test plan

- New test cases added


illicitonion (approved):
Seems reasonable! Thanks!

PR #521, opened 2025-04-14
chore: add config for Aspect Workflows

alexeagle:
Will replace our GitHub Actions CI, so we get persistent CI runners, buildbarn cache/RBE, and a web UI

### DIFF
--- a/.aspect/workflows/.bazelrc
+++ b/.aspect/workflows/.bazelrc
@@ -1,0 +1,6 @@
+# build without the bytes
+common --remote_download_outputs=minimal
+common --nobuild_runfile_links
+
+# Allows tests to run bazelisk-in-bazel, since this is the cache folder used
+common --test_env=XDG_CACHE_HOME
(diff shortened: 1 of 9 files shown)


kormide (approved):

alexeagle:
@kormide I guess I can leave GHA here just for testing the `example` folder with bzlmod disabled. Do you think it's easy to configure Workflows to test the `example` folder under both "matrix" jobs of bzlmod true/false? (Same question is going to come up in other repos for the matrix of testing with multiple Bazel versions)

kormide:
> @kormide I guess I can leave GHA here just for testing the `example` folder with bzlmod disabled. Do you think it's easy to configure Workflows to test the `example` folder under both "matrix" jobs of bzlmod true/false? (Same question is going to come up in other repos for the matrix of testing with multiple Bazel versions)

You could try creating two workspaces pointing to the same path with different `--enable_bzlmod` flag settings, but that's a path we haven't tested before and I'm unsure if it will work.

PR #525, opened 2025-04-15
Change close to work with fs sync methods

### LINKED ISSUE #524: [Bug]: Error: TypeError: diffOut.close is not a function
### What happened?

This is a bug induced from https://github.com/aspect-build/rules_lint/pull/511/files
Very sorry about it as I was the contributor.

@nlou9 brought this to my attention that the status code seemed to be not emitted.

I'm a bit confused as I tested this on our kafka repository; the different being it has **multiple aspects** so there may be another bug here of the status code being squashed (need to investigate)

https://github.com/aspect-build/rules_lint/pull/511/files#diff-b4fe27048a6cc502bc569e1b5c2820a154935be0fecd8e5990702b89b543c63aR111 is the culprit.

fzakaria:
openSync returns an fd and not an object so the call to `close` on it causes a failure.
Change to according closeSync call

fixes #524

### DIFF
--- a/lint/private/patcher.mjs
+++ b/lint/private/patcher.mjs
@@ -108,7 +108,7 @@ async function main(args, sandbox) {
     }
   }
 
-  diffOut.close();
+  fs.closeSync(diffOut);
 
   return ret.status;
 }


alexeagle (approved):
thanks!

PR #543, opened 2025-05-14
refactor: simplify devDep patching

alexeagle:
We can just use git archive stamping instead of managing and verifying a patch file


MODULE.bazel
```diff
@@ -39,8 +39,11 @@ register_toolchains("@sarif_parser_toolchains//:all")
 ####### Dev dependencies ########
 
 # To allow /tools to be built from source
-# NOTE: when publishing to BCR, we patch this to be True, as we publish pre-built binaries with our releases.
-IS_RELEASE = False
+# NB: this is stamped by git archive, see .gitattributes
+_VERSION_PRIVATE = "$Format:%(describe:tags=true)$"
+
+# If the version is like v1.2.3 but NOT like v2.0.3-7-g57bfe2c1, then there are pre-built binaries
+IS_RELEASE = _VERSION_PRIVATE.startswith("v") and _VERSION_PRIVATE.find("g") < 0
```
kormide, line 46: Is the `g` check reliable?

kormide (approved):

MODULE.bazel
alexeagle: I believe it is, that's what we've used forever in bazel-lib 
https://github.com/bazel-contrib/bazel-lib/blob/main/tools/version.bzl#L9-L15

Issue #552, opened 2025-05-22
Document where to find mypy integration


calliecameron:
The [type checkers section](https://github.com/aspect-build/rules_lint/tree/main#what-about-type-checking) in the readme says that mypy integration should be in rules_python or rules_py, but it isn't - it's actually in [rules_mypy](https://github.com/theoremlp/rules_mypy), which I only found by digging through comments in https://github.com/aspect-build/rules_lint/issues/79.

I know it's not really this repo's responsibility to link to other tools, but it would be convenient since they're often used alongside linters and formatters.

PR #553, opened 2025-05-22
Document where to find mypy.

calliecameron:
Document where to find mypy. Since type checkers are commonly used alongside linters and formatters, it's helpful to link to them, even when they're in other repos.

Fixes https://github.com/aspect-build/rules_lint/issues/552.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,34 +29,34 @@ Linters which are not language-specific:
 
 - [keep-sorted]
 
-| Language               | Formatter                 | Linter(s)                        |
-| ---------------------- | ------------------------- | -------------------------------- |
-| C / C++                | [clang-format]            | [clang-tidy]                     |
-| Cuda                   | [clang-format]            |                                  |
-| CSS, Less, Sass        | [Prettier]                | [Stylelint]                      |
-| Go                     | [gofmt] or [gofumpt]      |                                  |
-| Gherkin                | [prettier-plugin-gherkin] |                                  |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,34 +29,34 @@ Linters which are not language-specific:
...
+| HTML                   | [Prettier]                |                                  |                                |
+| JSON                   | [Prettier]                |                                  |                                |
+| Java                   | [google-java-format]      | [pmd] , [Checkstyle], [Spotbugs] |                                |
+| JavaScript             | [Prettier]                | [ESLint]                         |                                |
+| Jsonnet                | [jsonnetfmt]              |                                  |                                |
+| Kotlin                 | [ktfmt]                   | [ktlint]                         |                                |
+| Markdown               | [Prettier]                | [Vale]                           |                                |
+| Protocol Buffer        | [buf]                     | [buf lint]                       |                                |
+| Python                 | [ruff]                    | [flake8], [ruff]                 | [mypy]                         |
```
alexeagle, line 49: I think this makes the table very wide, and can't imagine any other rows that will ever have a value in the new column. Maybe we could treat it more like keep-sorted above, where it's an exception?  or just edit the FAQ at the bottom to point to mypy?

calliecameron: Done, added it in the FAQ.

alexeagle (approved):
awesome, thanks

Issue #549, opened 2025-05-22
[Bug]: Shell formatter doesn't find files with a shell shebang but no file extension


calliecameron:
### What happened?

The [shfmt docs](https://github.com/mvdan/sh/blob/master/cmd/shfmt/shfmt.1.scd) say 'EditorConfig sections may also use [[shell]] or [[bash]] to match any shell or bash scripts, which is particularly useful when scripts use a shebang but no extension'. However, even with this in .editorconfig, format_multirun doesn't format files with a shell shebang but no extension. Looks like this is because format.sh is only matching on file extension.

I've only seen this happen with shell files, but it would also affect e.g. python scripts with a shebang and no extension.

Ideally formatters should also look for shebangs, as well as any other in-file content common used to identify languages.

### Version

Development (host) and target OS/architectures: linux x64

Output of `bazel --version`: 8.2.1

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: 1.4.2

Language(s) and/or frameworks involved: shell, but also any language that uses extensionless files with shebangs

### How to reproduce

```shell

```

calliecameron:
Shfmt defines their shebang regex [here](https://github.com/mvdan/sh/blob/93f7ac212390b77f75b360440e35a22a9a09dc48/fileutil/file.go#L15).

calliecameron:
Actually it might be more reliable to use tools' own file discovery where possible, rather than regexes and git ls-files. Then users can configure discovery as usual in the formatter config file, there's no chance of mismatch between rules_lint and other runs, and you don't have to maintain regexes for extensions/shebangs.

alexeagle:
Yes, at least for the case of scanning the first line(s) of a file for a shebang line, this makes sense as an alternative to file extensions. I wonder how slow this will be though, if we do that for all files? Maybe it should only happen for files with no extension?

calliecameron:
Only doing it for files with no extension makes sense, I think. 

In the meantime I'm going to rename my extensionless files to have extensions, so the formatter picks them up, and think I might send another PR to document that extensionless files will be ignored.

psalaberria002:
I hit the same issue while trying to get our tools/bazel script formatted.

Synss:
I'm interested in fixing this issue. I'll start working on it now.

PR #556, opened 2025-05-23
Document that files with no extension will not be formatted

calliecameron:
Documenting the current behaviour of file discovery for formatters.

Part of https://github.com/aspect-build/rules_lint/issues/549.

### DIFF
--- a/docs/formatting.md
+++ b/docs/formatting.md
@@ -46,6 +46,8 @@ format_multirun(
 )
 ```
 
+File discovery for each language is based on file extension - files with no extension will not be formatted.
+
 ## Usage
 
 ### Configuring formatters


docs/formatting.md
```diff
@@ -46,6 +46,8 @@ format_multirun(
 )
 ```
 
+File discovery for each language is based on file extension - files with no extension will not be formatted.
```
alexeagle, line 49: could you add a link to the discussion about fixing this by sniffing the shebang line, just to make it discoverable for anyone who reads this and has the same problem you encountered?

calliecameron: Done.

Issue #551, opened 2025-05-22
[FR]: Add aliases for built-in lint tools


calliecameron:
### What is the current behavior?

Builtin formatter binaries using multitool are aliased in format/BUILD.bazel. It would be convenient if the multitool linters has similar aliases in lint/BUILD.bazel.

Currently, to use them, you have to add multitool as a direct dependency of your own module. It's not obvious that doing this would let you access the tools defined by rules_lint. This should be documented.

### Describe the feature

It would be convenient to have aliases for the linters, too.

alexeagle:
Sorry, I'm not quite following. Could you give some rough example of what you'd like it to look like, or how you'd like to use it?

calliecameron:
Or to put it another way: it's not obvious which tools are provided by rules_lint and how to use them, vs which ones you have to provide yourself.

calliecameron:
Sent PRs to update the docs. Now that I see that there are only two built-in tools, and it's clear how to use them, aliases maybe aren't required. Though it still seems a bit weird that the tools installed by rules_lint's multitool leak into the global multitool namespace - and aliases would avoid having to modify MODULE.bazel.

calliecameron:
I've been digging into this a bit more, and turns out I was confused about how this was working. The binary passed to `lint_shellcheck_aspect` and `lint_ruff_aspect` is resolved relative to rules_lint, _not_ relative to the user's module, because the string is being converted to a label inside the lint_*_macro (config being passed as a label should have made me think binary worked this way too). Which means my documentation updates are wrong. The user module doesn't need a direct dependency on multitool.

i.e. this:

```starlark
shellcheck = lint_shellcheck_aspect(
    binary = "@multitool//tools/shellcheck",  # uses the multitool dep of aspect_rules_lint
    config = Label("//:.shellcheckrc"),
)
```

is not the same as this:

```starlark
shellcheck = lint_shellcheck_aspect(
    binary = Label("@multitool//tools/shellcheck"),  # uses the multitool dep of the user's module
    config = Label("//:.shellcheckrc"),
)
```

The examples for all the other linters show binaries being passed using `@@` absolute labels, equivalent to `Label`, but even before my change, the examples for shellcheck and ruff showed binaries being passed using `@` relative labels, which confused me.

So I'm going to do a few things:
- add the aliases as before, and recommend using them in the docs with `Label`, so that it's obvious where they're coming from
- fix the shellcheck and ruff docs not to have the user depend on multitool
[...]

PR #557, opened 2025-05-23
Clarify how to use built-in formatters

calliecameron:
Also fixes the BUILTIN_TOOL_LABELS entry for ruff so it shows up in the docs.


alexeagle (approved):
Thanks!

PR #558, opened 2025-05-23
Clarify how to use built-in linters

calliecameron:

### DIFF
--- a/lint/ruff.bzl
+++ b/lint/ruff.bzl
@@ -14,11 +14,11 @@ use_repo(multitool, "multitool")
 Then create the linter aspect, typically in `tools/lint/linters.bzl`:
 
 ```starlark
-load("@aspect_rules_lint//lint:ruff.bzl", "ruff_aspect")
+load("@aspect_rules_lint//lint:ruff.bzl", "lint_ruff_aspect")
 
-ruff = ruff_aspect(
+ruff = lint_ruff_aspect(
     binary = "@multitool//tools/ruff",
-    configs = "@@//:.ruff.toml",
+    configs = ["@@//:.ruff.toml"],
 )
 ```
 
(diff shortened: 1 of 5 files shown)


lint/ruff.bzl
```diff
@@ -1,13 +1,24 @@
-"""API for declaring a Ruff lint aspect that visits py_library rules.
+"""API for declaring a Ruff lint aspect that visits py_{binary|library|test} rules.
 
 Typical usage:
 
+Ruff is provided as a built-in tool by rules_lint. To use the built-in version, first add a dependency on rules_multitool to MODULE.bazel:
```
alexeagle, line 5: drive-by question since you're current on this code, I wonder if we can remove the https://github.com/aspect-build/rules_lint/blob/main/lint/ruff_versions.bzl file

lint/ruff.bzl
```diff
@@ -1,13 +1,24 @@
...
-ruff = ruff_aspect(
+Then create the linter aspect, typically in `tools/lint/linters.bzl`:
+
+```starlark
+load("@aspect_rules_lint//lint:ruff.bzl", "lint_ruff_aspect")
+
+ruff = lint_ruff_aspect(
     binary = "@multitool//tools/ruff",
-    configs = "@@//:.ruff.toml",
+    configs = ["@@//:.ruff.toml"],
```
alexeagle, line 21: nit: we should stop using the `@@//some:target` and instead use explicit `Label(//some:target)`

alexeagle (approved):
thanks!

lint/ruff.bzl
calliecameron: Looks like it's only used by the workspace version of the rules, [here](https://github.com/aspect-build/rules_lint/blob/2dce2664821535ff581ebe6c93022f8273cda997/lint/ruff.bzl#L273). Depends whether you want to keep workspace support.

lint/ruff.bzl
calliecameron: Done.

Issue #559, opened 2025-05-23
Add a way to lint pyi files


calliecameron:
I'd like to lint my pyi files, but currently it doesn't look like there's a way to do it. lint_ruff_aspect only looks at py_binary, py_library and py_test targets, but they can only take .py files as srcs, not .pyi files.

Looks like the easiest way to do this would be to make the aspect also look at filegroups tagged 'python', like the vale linter does for markdown.

This would have the added benefit of being able to lint files that don't really belong in a py_* target, e.g. scripts intended to be run outside bazel, without having to create a dummy py_* rule for them - they could just go in a tagged filegroup instead. I'd be inclined to add this to shellcheck as well, for the same reason.

alexeagle:
Hm, is it even correct that `.pyi` files can't be sources of any `py_*` rules? @arrdem do you already know how someone is expected to add type-check-only inputs to the dependency graph?

PR #560, opened 2025-05-23
Allow the ruff linter to run on filegroups tagged 'python'

calliecameron:
This is for e.g. pyi files, which can't be in the srcs of a py_* rule. Also added the tag 'lint-with-ruff' for consistency with the vale linter.

Part of https://github.com/aspect-build/rules_lint/issues/559.

### DIFF
--- a/example/src/BUILD.bazel
+++ b/example/src/BUILD.bazel
@@ -66,6 +66,12 @@ py_library(
     srcs = ["unused_import.py"],
 )
 
+filegroup(
+    name = "unused_import_filegroup",
+    srcs = ["unused_import.py"],
+    tags = ["python"],
+)
+
 java_library(
     name = "foo",
     srcs = ["Foo.java"],
--- a/example/test/BUILD.bazel
+++ b/example/test/BUILD.bazel
@@ -52,9 +52,21 @@ py_library(
     srcs = ["generated.py"],
 )
 
+filegroup(
+    name = "generated_py_filegroup",
+    srcs = ["generated.py"],
+    tags = ["lint-with-ruff"],
+)
+
 py_library(
     name = "excluded_py",
     srcs = ["excluded.py"],
+)
+
+filegroup(
+    name = "excluded_py_filegroup",
+    srcs = ["excluded.py"],
+    tags = ["python"],
 )
 
 java_library(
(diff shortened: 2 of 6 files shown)


example/src/BUILD.bazel
```diff
@@ -66,6 +66,12 @@ py_library(
     srcs = ["unused_import.py"],
 )
 
+filegroup(
+    name = "unused_import_filegroup",
+    srcs = ["unused_import.py"],
```
alexeagle, line 71: since your stated motivation is to allow `.pyi` files, could you change the example to illustrate that? I don't want this file to be confusing to users. Also having an input in the `srcs` of two neighboring rules isn't a pattern we want users to copy.

lint/ruff.bzl
```diff
@@ -1,4 +1,4 @@
-"""API for declaring a Ruff lint aspect that visits py_library rules.
+"""API for declaring a Ruff lint aspect that visits py_library rules and filegroups tagged 'python'.
```
alexeagle, line 1: I'd omit this edge case from the very short overview doc

lint/ruff.bzl
```diff
@@ -187,13 +187,14 @@ def _ruff_aspect_impl(target, ctx):
...
 
-def lint_ruff_aspect(binary, configs, rule_kinds = ["py_binary", "py_library", "py_test"]):
+def lint_ruff_aspect(binary, configs, rule_kinds = ["py_binary", "py_library", "py_test"], filegroup_tags = ["python", "lint-with-ruff"]):
     """A factory function to create a linter aspect.
 
     Attrs:
         binary: a ruff executable
         configs: ruff config file(s) (`pyproject.toml`, `ruff.toml`, or `.ruff.toml`)
         rule_kinds: which [kinds](https://bazel.build/query/language#kind) of rules should be visited by the aspect
+        filegroup_tags: filegroups tagged with these tags will be visited by the aspect
```
alexeagle, line 197: add "in addition to Python rule kinds"

example/test/BUILD.bazel
```diff
@@ -78,11 +90,21 @@ ruff_test(
...
+    name = "ruff_should_ignore_generated_filegroup",
+    srcs = [":generated_py_filegroup"],
+)
+
 ruff_test(
     name = "ruff_should_ignore_excluded",
     srcs = ["excluded_py"],
 )
 
+ruff_test(
```
alexeagle, line 103: it feels like a lot more lines of tests are added here than are warranted to illustrate and verify the small code change, could you distill it down a bit? (I think a single filegroup is sufficient to get the needed "test coverage")

alexeagle (commented):
LGTM with some fixes - thanks!

lint/ruff.bzl
calliecameron: Done.

lint/ruff.bzl
calliecameron: Done.

alexeagle (approved):

PR #563, opened 2025-05-29
Add aliases for built-in linters

calliecameron:
These are equivalent to the aliases for the built-in formatters, and allow use of these linters without having to use confusing relative repo names that appear to be relative to the user's module, but are actually relative to rules_lint.

Ideally the aliases would be `ruff` and `shellcheck`, but those names are taken by bzl_libraries, hence the 'bin' suffix.

### DIFF
--- a/docs/ruff.md
+++ b/docs/ruff.md
@@ -4,22 +4,14 @@ API for declaring a Ruff lint aspect that visits py_{binary|library|test} rules.
 
 Typical usage:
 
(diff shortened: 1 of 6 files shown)


calliecameron:
@alexeagle any update on reviewing this? This is the last PR for https://github.com/aspect-build/rules_lint/issues/551, and also contains the last of the related doc updates.

alexeagle (approved):
thanks! Sorry I missed the notifications on this one

PR #571, opened 2025-06-13
fix: make correct bzl_library targets

alexeagle:
See https://github.com/bazelbuild/bazel-skylib/issues/568


jbedard:
Can this be reverted once https://github.com/bazelbuild/bazel-skylib/issues/568 is fixed? Can we put a link to https://github.com/bazelbuild/bazel-skylib/issues/568 in a comment somewhere?

jbedard (approved):

PR #576, opened 2025-06-24
feat: Format toml with taplo

Synss:
This patch adds formatting for [TOML](https://toml.io/en/) files. TOML is often used to configure various tooling and languages
(for example, `.ruff.toml`, `pyproject.toml`, `Cargo.toml`).

I chose to use [taplo](https://taplo.tamasfe.dev/) directly as the formatter, although my first attempt used [prettier-plugin-toml](https://www.npmjs.com/package/prettier-plugin-toml). My company isn't very js-focused so that obtaining `taplo` directly is much simpler than going over `prettier`.

`taplo` has some linting features, not implemented here, as we don't use them where I work and I don't know them.

---

### Changes are visible to end-users: yes, new formatter! 😊

- Searched for relevant documentation and updated as needed: yes (`README.md`)

### Test plan

- New test cases added
- I also readily patched our own repo with this change and let `bazel run format` run. It did reformat some of our `toml` files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ Linters which are not language-specific:
 | Shell                  | [shfmt]                   | [shellcheck]                     |
 | Starlark               | [Buildifier]              |                                  |
 | Swift                  | [SwiftFormat] (1)         |                                  |
+| TOML                   | [taplo]                   |                                  |
 | TSX                    | [Prettier]                | [ESLint]                         |
 | TypeScript             | [Prettier]                | [ESLint]                         |
 | YAML                   | [yamlfmt]                 |                                  |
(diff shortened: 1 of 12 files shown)


Synss:
I don't understand why the CI is red. Should I copy/move my `http_file` call from the `MODULE.bazel` file to the `WORKSPACE.bazel` file?

alexeagle:
Yes, this repo still supports Bazel users who haven't switched to bzlmod, so the dependency has to be copied to https://github.com/aspect-build/rules_lint/blob/main/format/repositories.bzl#L18

Synss:
@alexeagle : Thank you for your quick answer! 🟢, finally.

alexeagle (approved):
Thanks!

PR #597, opened 2025-08-13
Improve support of hemertic toolchains with the clang tidy linter

Synss:
The changes are more detailed in the commit messages.

In short, this adds two optional arguments to the linter:
 * `gcc-install-dir` arg to prevent `clang-tidy` to use a gcc intallation from the host.
 * `deps` lets users actually provide the headers and files from an hermetic toolchain registered elsewhere using the `rules_cc` API.

both options were lifted from https://github.com/erenon/bazel_clang_tidy

### Test plan

Would require registering an hermetic gcc toolchain, which is not trivial.

But manually:
* register hermetic gcc toolchain (based on `x86-64--glibc--stable-2024.05-1.tar.xz` from https://bootlin.com/ for example, see also https://github.com/bazelbuild/rules_cc/tree/main/examples/rule_based_toolchain/toolchains/gcc)
* set `gcc_install_dir` and `deps` arguments to the label defined as follows (using the bootlin toolchain)
```
directory(
    name = "x86_64-buildroot-linux-gnu",
    srcs = ["lib/gcc/x86_64-buildroot-linux-gnu/13.3.0"],
)
```
* make sure that `clang-tidy` now uses the hermetic headers.

### DIFF
--- a/lint/clang_tidy.bzl
+++ b/lint/clang_tidy.bzl
@@ -51,6 +51,8 @@ def _gather_inputs(ctx, compilation_context, srcs):
     inputs = srcs + ctx.files._configs
     if (any(ctx.files._global_config)):
         inputs.append(ctx.files._global_config[0])
+    for dep in ctx.files._deps:
+        inputs.append(dep)
     return depset(inputs, transitive = [compilation_context.headers])
 
 def _toolchain_env(ctx, user_flags, action_name = ACTION_NAMES.cpp_compile):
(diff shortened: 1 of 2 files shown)


lint/clang_tidy.bzl
```diff
@@ -37,6 +37,7 @@ clang_tidy = lint_clang_tidy_aspect(
 ```
 """
 
+load("@bazel_skylib//rules/directory:providers.bzl", "DirectoryInfo")
```
alexeagle: ooh #TIL

PR #602, opened 2025-08-20
fix: include ts_project(tsconfig) in eslint inputs

jbedard:
eslint plugins often need the tsconfig

Include `ts_project(tsconfig)` as inputs to eslint aspect.

### Test plan


alexeagle (approved):
ideally we'd add an example as a test that shows it's load-bearing

PR #605, opened 2025-08-21
windows formatters

peakschris:
Update the toolchain to include windows builds of each formatter.

These missing windows builds are preventing windows tests from running in rules_js


peakschris:
@alexeagle could you please help me to understand how to update rules_multitool to 0.11.0 in the two test environments? This is why the two CI tests are failing.

alexeagle:
Added a commit to green it up.

PR #625, opened 2025-09-30
chore: fix markdown parsing issue

stagnation:
A simple typo broke the parsing and rendering of the tools table


alexeagle (approved):
thanks!

PR #636, opened 2025-10-06
chore: Format '*.MODULE.bazel' files

honnix:
To format https://bazel.build/rules/lib/globals/module#include files.

---

This change also include a new file pattern for starlark file discovery: `*.MODULE.bazel`, which is for included files as documented by https://bazel.build/rules/lib/globals/module#include

### Test plan

We have been using this patch for a while.


honnix:
CI failure doesn't seem to be caused by this change.

alexeagle (approved):

PR #649, opened 2025-10-28
fix: combine copts and cxxopts for improved compiler argument handling

mutalibmohammed:
This pull request updates how compiler flags are collected in the `_get_compiler_args` function. The change ensures that both `copts` and `cxxopts` attributes are included when gathering rule-specific flags.

* Now, both `copts` and `cxxopts` attributes are combined and used as rule flags in the `_get_compiler_args` function in `lint/clang_tidy.bzl`.


mutalibmohammed:
@alexeagle Could I have a review for this please?

PR #650, opened 2025-10-29
feat: Add formatting support for C# and F#

purkhusid:
## What?
Adds formatting support for F# and C# using fantomas(F#) and CSharpier(C#)

The formatters are not avilable as standalone executables and are provided as NuGet packages via rules_dotnet by the end-user for rules_lint

---

### Test plan

- New test cases added

### DIFF
--- a/example/WORKSPACE.bazel
+++ b/example/WORKSPACE.bazel
@@ -411,10 +411,6 @@ load("@aspect_rules_lint//lint:spotbugs.bzl", "fetch_spotbugs")
 
 fetch_spotbugs()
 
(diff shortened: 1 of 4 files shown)


purkhusid:
@alexeagle I'm not sure what to do about WORKSPACE here. I removed support for WORKSPACE in rules_dotnet a while back.

alexeagle:
Can you add some conditional to disable them? 
`is_bzlmod_enabled = str(Label("//:invalid")).startswith("@@")` is an easy way to detect the flag

purkhusid:
@alexeagle Everything is green now

PR #661, opened 2025-11-23
chore: manual mirror ruff releases

hunshcn:
Since mirror_ruff.sh can only mirror the latest release, we need to do a manual mirror before https://github.com/aspect-build/rules_lint/pull/660 merge.


alexeagle:
I've got to figure out why CI is red, but it's no worse on main, hopefully ruff will be green again at the end. Thanks!

PR #664, opened 2025-11-27
chore: manual mirror external releases

hunshcn:
Because of github action is still broken, there is a manual update.

This has also adapted to the changes of artifact name in https://github.com/google/go-jsonnet/releases/tag/v0.21.0

### DIFF
--- a/format/multitool.lock.json
+++ b/format/multitool.lock.json
@@ -4,36 +4,36 @@
     "binaries": [
       {
         "kind": "file",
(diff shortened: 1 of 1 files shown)


hunshcn:
gofumpt include a breaking change in v0.9.1 and rollback on v0.9.2 https://github.com/mvdan/gofumpt/issues/285

we need this update

hunshcn:
PTAL @alexeagle

Issue #385, opened 2024-09-10
[FR]: Support clippy from rules_rust


mrmeku:
### What is the current behavior?

Currently linting for rust is not supported by rules_lint. Rules_rust can run clippy, the rust linter, but it follows different idioms than rules_lint.

### Describe the feature

Utilize the existing aspect from rules_rust to enable running clippy with rules_lint

To run clippy the rules_rust way you would augment your `.bazelrc` file with
```
# https://bazelbuild.github.io/rules_rust/rust_clippy.html
build --aspects=@rules_rust//rust:defs.bzl%rust_clippy_aspect
build --output_groups=+clippy_checks
build --@rules_rust//:clippy.toml=//:clippy.toml
```

We want to create a rules_lint shapped aspect that can be configured by
```
load("@aspect_rules_lint//lint:clippy.bzl", "lint_clippy_aspect")
eslint = lint_eslint_aspect(
    configs = "//:clippy.toml",
)
```

In so doing, we can enable rules_lint users to run clippy without augmenting their `.bazelrc` and leave requesting the `clippy_checks` output group to the aspect CLI or other mechanism to run rules_lint

blorente:
I’ve opened a PR to handle exit codes the right way: https://github.com/aspect-build/rules_lint/pull/673
It does depend on another PR from rules_rust though, so it’s not mergeable yet.

PR #667, opened 2025-12-05
feat: Implement clippy support 

blorente:
Adds clippy support to rules_lint, which is part of #385. It does not support `--fix` yet, as there are some complications with applying the patches generated by rules_rust.

Caveats:
- `fix` doesn't work.
- The machine outputs are in in clippy's native JSON, not SARIF. We could use cargo-sarif to transform it, but we haven't for now.
- We treat all clippy warnings as errors. This is because `rules_rust` will fail the entire execution if there's even one error, so we need to limit the reports to just warnings so that we can continue the target execution and generate useful output files. Because we limit all errors to warnings, we must consider every warning as an error.

---

- Breaking change (forces users to change their own code or config): Only if they were depending on rules_rust before, it may change their `MODULE.lock` resolution.

### Test plan

- New test cases added
- Some of the cases are manual: `cd examples && bazel test //test:clippy_ok_binary_with_bad_dep //test:clippy_bad_binary //test:clippy_bad_library`

### DIFF
--- a/example/WORKSPACE.bazel
+++ b/example/WORKSPACE.bazel
@@ -134,6 +134,25 @@ load("@bazel_features//:deps.bzl", "bazel_features_deps")
 
 bazel_features_deps()
 
+load("@jq.bzl//jq/toolchain:platforms.bzl", "JQ_PLATFORMS", "jq_platform_repo")
+load("@jq.bzl//jq/toolchain:toolchain.bzl", "DEFAULT_JQ_VERSION", "jq_toolchains_repo")
+
+[
+    jq_platform_repo(
+        name = "{}_{}".format("jq_toolchains", platform),
+        platform = platform,
+        version = DEFAULT_JQ_VERSION,
+    )
+    for platform in JQ_PLATFORMS.keys()
+]
+
+jq_toolchains_repo(
+    name = "jq_toolchains",
+    user_repository_name = "jq_toolchains",
+)
+
+register_toolchains("@jq_toolchains//:all")
+
 load("@yq.bzl//yq/toolchain:platforms.bzl", "YQ_PLATFORMS", "yq_platform_repo")
 load("@yq.bzl//yq/toolchain:versions.bzl", "DEFAULT_YQ_VERSION")
 


lint/clippy.bzl
```diff
@@ -0,0 +1,185 @@
+"""API for declaring a clippy lint aspect that visits rust_{binary|library|test} rules.
+
+Typical usage:
+
+First, install `rules_rust` into your repository: https://bazelbuild.github.io/rules_rust/. For instance:
```
alexeagle: does it require some minimum version of rules_rust to pick up your changes? would be good to document

MODULE.bazel
```diff
@@ -24,6 +24,9 @@ bazel_dep(name = "rules_multitool", version = "0.11.0")
 bazel_dep(name = "rules_diff", version = "1.0.0")
 bazel_dep(name = "rules_shell", version = "0.5.0")
 
+# Needed in the root because we use the run_clippy action in our aspect impl
+bazel_dep(name = "rules_rust", version = "0.67.0")
```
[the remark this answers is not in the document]
alexeagle: Thanks for studying that. I agree there isn't a better option right now. We're adding a `load` to rules_rust in our `bzl` code so we have to dep on it. Even if we vendored some rules_rust code we'd still need to reference providers so I think it's unavoidable.

lint/clippy.bzl
```diff
@@ -0,0 +1,191 @@
...
+        config = ctx.file._config_file,
+        output = outputs.machine.out,
+        success_marker = machine_success_indicator,
+        cap_at_warnings = True,
+        extra_clippy_flags = extra_options,
+        error_format = "json",
+    )
+    _marker_to_exit_code(ctx, machine_success_indicator, outputs.machine.out, outputs.machine.exit_code)
+
+    # FIXME: Rustc only gives us JSON output, which we can't turn into SARIF yet.
```
alexeagle: In theory https://github.com/reviewdog/errorformat/blob/master/fmts/rust.go should guide us to the answer

alexeagle (approved):

PR #669, opened 2025-12-09
fix: Add Stylelint stdout to error output file

tomsoir:
This change allows for displaying StyleLint outputs when it exits with 0 code. 

TL, DR: The reason why eslint currently works correctly (prints out eslint outputs) is that it uses the standard output of the command stdout (JS_BINARY__STDOUT_OUTPUT_FILE [env](https://github.com/aspect-build/rules_lint/blob/main/lint/eslint.bzl#L199)), while the current Stylelint implementation only uses the stderr (JS_BINARY__STDERR_OUTPUT_FILE [env](https://github.com/aspect-build/rules_lint/blob/main/lint/stylelint.bzl#L160)), which works well when Stylelint itself terminates with critical errors BUT it remains empty when it terminates with a zero exit code (e.g. when --fix parameter is enabled —> it exits with 0 exit code but still producers outputs/stylelint results that were found/fixed).

---

### Test plan

- New test cases added


alexeagle (approved):
thanks for putting in the effort here!

https://github.com/aspect-build/rules_lint/blob/main/lint/eslint.bzl#L199 - yeah this is just making it consistent 👍🏻

PR #671, opened 2025-12-10
refactor: remove content of WORKSPACE for bzlmod example

alexeagle:
Introduce a module extension to provide tool repos.
Allows upgrade to Bazel 9

Note, this introduces duplication with a bunch of repo-macros called `fetch_*` which I'll remove in a 2.x branch

### DIFF
--- a/example/tools/lint/BUILD.bazel
+++ b/example/tools/lint/BUILD.bazel
@@ -4,8 +4,8 @@ This is in its own package because it has so many loading-time symbols,
 we don't want to trigger eager fetches of these for builds that don't want to run format.
 """
 
-load("@bazel_lib//lib:copy_to_directory.bzl", "copy_to_directory")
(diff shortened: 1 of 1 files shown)


example/MODULE.bazel
```diff
@@ -182,23 +172,39 @@ use_repo(ruby, "bundle", "ruby", "ruby_toolchains")
...
+    "com_github_pinterest_ktlint",
+    "com_puppycrawl_tools_checkstyle",
+    "cppcheck_linux",
+    "cppcheck_macos",
+    "net_sourceforge_pmd",
+    "spotbugs",
+    "vale_Google",
+    "vale_Linux_64-bit",
+    "vale_Linux_arm64",
+    "vale_Windows_64-bit",
```
jbedard: The mixed case binaries 😨

alexeagle: yeah well... that's existing and I don't think it's bad enough to break any users

PR #688, opened 2025-12-27
fix(ty): check existence of extra search paths before passing to ty

protocol7:
When using `lint_ty_aspect` from rules_lint with Python targets that depend on pip packages, the ty linter fails because some pip package directories referenced in `--extra-search-path` don't exist, causing ty to error out.

The `ty.bzl` implementation collects import paths from `PyInfo.imports` and unconditionally passes them all to ty via `--extra-search-path`. However, not all directories in `PyInfo.imports` actually exist. Some pip packages may have transitive dependencies whose directories are missing in the build. `ruamel.yaml` is one example of such a package.

This change modifies the action to check if each directory exists before adding it to `--extra-search-path`.

### DIFF
--- a/examples/python/BUILD
+++ b/examples/python/BUILD
@@ -11,6 +11,8 @@ exports_files(
     visibility = ["//visibility:public"],
 )
 
+# Update requirements.txt by running:
+# bazel run //:requirements.update
 pip_compile(
     name = "requirements",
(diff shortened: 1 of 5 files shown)


alexeagle (approved):
I added a reproduction of the issue to your PR.
Thanks!

PR #701, opened 2026-01-07
feat: Add semgrep linter

Synss:
This patch adds support for [semgrep](https://semgrep.dev/).

---

### Test plan

- New test cases added
- Manual testing (in work repo).


Synss:
Linting every file separately like this is too slow in the CI. We'll most likely drop the tool instead.

PR #707, opened 2026-01-11
fix(clippy): Store exit code to respect errors

blorente:
Assume[ this change in rules_rust](https://github.com/bazelbuild/rules_rust/pull/3772) is going to merge, and take advantage of it to keep the clippy exit code in a file and avoid differentiating between warnings and errors. 

- Suggested release notes appear below: no (clippy is not released yet, so we don't include notes yet)

### Test plan

- New test cases added (but manual)


alexeagle (approved):

alexeagle (approved):

alexeagle (approved):
🌮

PR #708, opened 2026-01-11
feat(clippy): Add SARIF conversion support

blorente:
Clean up support for SARIF parsing in `//lint/rust`, and use it to transform the output of clippy into SARIF JSON reports.
Depends on #707 

- Breaking change (forces users to change their own code or config): no (unless they were relying on the machine file to contain raw clippy diagnostics).

### Test plan

- New test cases added (and cleaned up the old ones)


blorente:
@alexeagle  Rebased and green :)

Issue #698, opened 2026-01-06
[Bug]: lint command in lint.axl hangs


alexeagle:
### What happened?

the lint command often hangs for me. Reproduces on CI as well:

https://github.com/aspect-build/rules_lint/blob/main/.github/workflows/ci.yaml#L79-L81

### Version

HEAD

PR #715, opened 2026-01-13
feat(clippy): Add support for `--fix`

blorente:
Depends on #708 .

Add support for --fix to the clippy linter.

Clippy prints diagnostics in the standard rustc diagnostic JSON format (https://doc.rust-lang.org/beta/rustc/json.html#diagnostics), which includes machine-readable instructions. Since `//lint/rust:cli` can already parse diagnostic files, we extend it to handle modifying files and writing them back to disk.

The implementation is heavily based on [`rustfix`](https://crates.io/crates/rustfix),, the official Rust crate to generate code fixes from diagnostics, particularly this disambiguation code: https://github.com/rust-lang/cargo/blob/master/crates/rustfix/src/replace.rs#L143-L192.

With that, we make the standard rules_lint patcher run `//lint/rust:cli` just like any other linter to generate diffs.

Note: To test this behaviour, we have introduced a new rule, `patcher_run`, which allows us to assert on patcher behaviour.

- Searched for relevant documentation and updated as needed: yes, but no updates needed as most linters don't advertise --fix compatibility in their documentation.
- Suggested release notes appear below: no, I'm not sure we want to write them yet.

### Test plan

- New test cases added

### DIFF
--- a/lint/private/patcher_run.bzl
+++ b/lint/private/patcher_run.bzl
@@ -29,7 +29,7 @@ def _patcher_run_impl(ctx):
         patch_cfg_env = dict(env, **{"BAZEL_BINDIR": bindir}),
         env = env,
         mnemonic = ctx.attr.mnemonic,
-        progress_message = "Running patcher in %{label}",  # TODO BL: This could be better.
+        progress_message = "Running patcher in %{label} as part of the build",
     )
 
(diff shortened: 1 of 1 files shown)


blorente:
@alexeagle I think the failing tests may be due to the hanging we're seeing in #698

alexeagle (approved):

PR #737, opened 2026-01-30
fix: Include *.bazel files except *.lock.bazel files.

nigelgbanks:
## Include `*.bazel` files in Starlark formatting

Adds `*.bazel` to the Starlark file patterns to format files like `foo.bazel`. Previously only specific files like `BUILD.bazel`, `MODULE.bazel`, and `WORKSPACE.bazel` were matched.

To avoid formatting generated lockfiles, this also introduces an exclusion mechanism that filters out `*.lock.bazel` files (e.g., `multitool.lock.bazel`).

---

**Release notes:** The Starlark formatter now includes `*.bazel` files. Lockfiles matching `*.lock.bazel` are excluded.

### Test plan

  1. Create a `foo.bazel` file and a `bar.lock.bazel` file in your workspace
  2. Run the format command targeting Starlark files
  3. Verify `foo.bazel` is formatted and `bar.lock.bazel` is excluded


alok-oai:
> `foo` in this case is just a [placeholder](https://en.wikipedia.org/wiki/Metasyntactic_variable). A more concrete example would be where you can have more than one `BUILD.bazel` file in a folder. Say when using [new_local_repository](https://bazel.build/rules/lib/repo/local#new_local_repository) and supplying a `BUILD.package.bazel` file to for `build_file`, that does the building for that repository. As well as having a `BUILD.bazel` file in the same folder for packaging a target from that repository with your application. Separating the packaging logic (unique to your application) from the repositories building logic (specific to that repository).

We also follow the convention of using BUILD.package.bazel for third-party repository overlays. Having these files be auto-formatted would be nice.

PR #746, opened 2026-02-04
fix(axl): exclude deleted files from formatting

rafikk:
The `format` aspect task uses `git diff --name-only MERGE_BASE` to detect changed files. However, by default, this command returns paths to deleted files. When trying to format these files, of course, the formatter fails.

This PR adds `--diff-filter=d` to the `git diff` command to filter out deleted files.


alexeagle (approved):
thanks!

PR #711, opened 2026-01-12
feat: Add scalafix linter support

greggdonovan:
Add Scalafix as a linter for Scala code. Scalafix is a refactoring and linting tool from Scala Center that supports both syntactic and semantic code analysis.

Features:
- Supports both syntactic and semantic modes
- Fix mode support via patcher_action
- SARIF output for CI integration

---

### Changes are visible to end-users: yes

- Searched for relevant documentation and updated as needed: yes
- Breaking change (forces users to change their own code or config): no
- Suggested release notes appear below: yes

### Suggested Release Notes

- Added support for [scalafix](https://scalacenter.github.io/scalafix/) as a linter. Requires `rules_scala` toolchain configuration for semanticdb support.

### Testing

Running the example:

```
scala ➤ aspect lint                                                                                                                   git:feat/scalafix-linter
Computing main repo mapping:
Loading:
Loading: 0 packages loaded
Analyzing: 10 targets (0 packages loaded, 0 targets configured)
Analyzing: 10 targets (0 packages loaded, 0 targets configured)
[0 / 1] checking cached actions
INFO: Analyzed 10 targets (0 packages loaded, 222 targets configured).
[4 / 50] [Prepa] Executing genrule @@rules_jvm_external++maven+maven//:ch_epfl_scala_scalafix_interfaces_0_14_5_extension [for tool] ... (16 actions, 1 running)
[86 / 88] [Prepa] Linting //src:semantic_test with Scalafix
/private/var/tmp/_bazel_gdonovan/bf7f14c4eaa63abf8c0b9a76fb5b1fd6/sandbox/darwin-sandbox/2009/execroot/_main/src/hello.scala:3:3: error: [DisableSyntax.var] mutable state should be avoided
  var mutableState = 0
  ^^^
/private/var/tmp/_bazel_gdonovan/bf7f14c4eaa63abf8c0b9a76fb5b1fd6/sandbox/darwin-sandbox/2009/execroot/_main/src/hello.scala:6:27: error: [DisableSyntax.null] null should be avoided, consider using Option instead
  val nullValue: String = null
                          ^^^^
[87 / 88] Linting //src:semantic_test with Scalafix; 1s darwin-sandbox
INFO: Found 10 targets...
INFO: Elapsed time: 4.437s, Critical Path: 2.19s
INFO: 44 processes: 45 action cache hit, 2 internal, 41 darwin-sandbox, 1 worker.
INFO: Build completed successfully, 44 total actions
```

PR #750, opened 2026-02-06
fix(sarif): pass full label to sarif tool

rafikk:
The SARIF processor has logic in `determineRelativePath` to produce source-relative paths from absolute paths. This logic depends on parsing the `--label` flag to identify the containing Bazel package.

However, `parse_to_sarif_action` only passes `ctx.label.name` as the value for `--label` which excludes the package name. This short-circuits the logic to produce proper paths. This PR updates the argument to `ctx.label` which passes full label names.

I detected this when testing #711, which produces SARIF output paths of the form `../../../2700/execroot/_main/jvm/com/foo/Foo.scala` for a target `//jvm/com/foo:foo`. With this patch applied, it produces the expected output `jvm/com/foo/Foo.scala`.

#### Suggested Release Notes

- Fixed a bug in SARIF output processing which may have produced incorrect file paths in reports.

### Test plan

Tested manually. There should probably be end-to-end tests for this as well. However, the tests cases in `sarif_test.go` all use absolute Bazel labels which this PR plumbs `parse_to_sarif_action` to provide.

### DIFF
--- a/lint/private/lint_aspect.bzl
+++ b/lint/private/lint_aspect.bzl
@@ -234,7 +234,7 @@ def parse_to_sarif_action(ctx, mnemonic, raw_machine_report, sarif_out):
     args = ctx.actions.args()
     args.add("-in", raw_machine_report.path)
     args.add("-out", sarif_out.path)
-    args.add("-label", ctx.label.name)
+    args.add("-label", ctx.label)
     args.add("-mnemonic", mnemonic)
 
     ctx.actions.run(


rafikk:
@alexeagle After testing a few more linters, I still only see this issue exhibited in the (in-progress) scalafix linter. However, this does appear to be a real bug. It seems like the code path in `determineRelativePath` just isn't in use by most linters.

Do you think we could merge this change?

Issue #763, opened 2026-02-12
[Bug]: ty lint aspect is very noisy on success


TrapsterDK:
### What happened?

Related to issue #753

When running ty lint it results in a lot of success messages like:
```
All checks passed!
All checks passed!
All checks passed!
```

This issue has already been solved and merged in for ruff with the PR #757, but did not fix ty.

The same patch can be applied to ty as far as I can see, with just adding --quiet as an argument which ty supports.

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`:
9.0.0

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:
bazel_dep(name = "aspect_rules_lint", version = "2.1.0")

Language(s) and/or frameworks involved:

### How to reproduce

```shell

```

TrapsterDK:
The fix for this issue might not be as straightforward as ty --quiet does not do the same as ruff. Instead it produces a result like this when it encounters errors `Found 5 issues.`, which is not what is wanted. A possible solution is to pipe the output and then only print it if it does not exit 0.

PR #777, opened 2026-02-27
feat(ty): Quiet on success

TrapsterDK:
Related to #763 

Made ty not print all checks passed on success which overflows command line.

---

- Searched for relevant documentation and updated as needed: not needed
- Breaking change (forces users to change their own code or config): no, but yeah if they depended on the output for every file
- Suggested release notes appear below: Ty quiet on success

### Test plan

Tested manually against larger code base. Should also be reproducible on code provided in examples/.

### DIFF
--- a/lint/ty.bzl
+++ b/lint/ty.bzl
@@ -92,17 +92,31 @@ fi
 """.format(path = path)
 
     color_flag = "--color always" if color else "--color never"
+    command = """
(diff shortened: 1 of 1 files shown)


alexeagle:
oh shoot @TrapsterDK do you mind rebasing over the merge conflict?

Issue #786, opened 2026-03-02
[Bug]: Clippy fails when rust target has generated inputs


purkhusid:
### What happened?

When a rust_library/rust_binary has a generated input the Clippy aspect in rules_lint fails due to conflicting generated files. 

Reproduction can be found here: https://github.com/purkhusid/rules_lint/tree/clippy-repro

I'm going to continue debugging this to find the root cause but just wanted to create the issue in case @blorente might know a obvious fix.

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: 9.0.0

rules_lint at HEAD

Language(s) and/or frameworks involved:
Rust

### How to reproduce

Repro here: https://github.com/purkhusid/rules_lint/tree/clippy-repro
Run `cd examples/rust && bazel build //... --@aspect_rules_lint//lint:fix` to get the following error:
```shell

ERROR: file 'src/ok_binary.rs' is generated by these conflicting actions:
Label: //src:ok_binary_with_generated_inputs
Aspects: [//tools/lint:linters.bzl%clippy], []
RuleClass: rust_binary rule
JavaActionClass: class com.google.devtools.build.lib.analysis.actions.StarlarkAction, class com.google.devtools.build.lib.analysis.actions.SymlinkAction
Configuration: 25295d47ca4f6ac508b9b60443fee8eb136d08978c375ba0944d73eeaac4c105
Mnemonic: CopyFile, Symlink
IsShareable: true
Action key: 79c0e22133eccc03c26a734724830c24b43bfeed90bf33fd989c0d639a6476d5, c90914cc1e7ea06f84518cf4f179e63dbd20dad9358a550644494828e4a4ed28
Progress message: Copying file src/ok_binary.rs, Creating symlink to source file: src/ok_binary.rs
Action describeKey: Copying file src/ok_binary.rs
  Execution info: no-sandbox=1
  Argument: external/bazel_lib++toolchains+coreutils_linux_amd64/coreutils
  Argument: cp
  Argument: src/ok_binary.rs
  Argument: bazel-out/k8-fastbuild/bin/src/ok_binary.rs
  Output paths mode: OFF
, GUID: 7f4fab4d-d0a7-4f0f-8649-1d0337a21fee
inputPath: null

PrimaryInput: File:[/home/danielp/dannidev/repos/purkhusid/rules_lint/examples/rust[source]]src/ok_binary.rs
[...]

PR #790, opened 2026-03-03
fix(clippy): Clippy autofix on targets that contain generated sources

purkhusid:
This is a fix for #786. 

The approach taken is to create the exact same action as is generated by rules_rust so that Bazel will deuplicate the actions.

The downside to this approach is that it relies on how rules_rust handles things internally. I'm also up for doing a more principled approach if there are any suggestions on how to do so.

### Changes are visible to end-users: no

### Test plan

Added a test that covers possible regression


purkhusid:
@blorente Since you created the original rules_rust integration.

blorente:
I don't have a better solution, happy to merge as-is.

PR #806, opened 2026-03-11
refactor(format): remove .jsh and .jav extensions for java

mattnworb:
google-java-format will only format files that end in `.java`, printing a warning for anything else: 

https://github.com/google/google-java-format/blob/2f7e558e586788110dfda603652f68915d890cb0/core/src/main/java/com/google/googlejavaformat/java/Main.java#L142-L145

while in theory someone using rules_lint can configure the formatter_multirun target to use another formatter binary, since GJF is the one advertised in the README I'm not sure it makes a lot of sense to default to passing it file extensions it will print warnings about.

---

### Changes are visible to end-users: maybe

- Breaking change (forces users to change their own code or config): yes, if they are using a different formatter that actually formats .jsh files and they have them in their repo


mattnworb:
> thanks, could you update the extension mirroring script too? I don't seem to have permission to push to your branch.

sorry @alexeagle I didn't see this until now! created #825 with that patch

PR #832, opened 2026-04-07
fix(ty): improve generated file handling

comicfans:
if dep contains generated source file, it will be placed under ctx.bin_path, for example generated source file from external module will be placed under `./bazel-out/k8-fastbuild/bin/external/ext_mod+/src/generated.py` and in-workspace ones will be placed under `./bazel-out/k8-fastbuild/bin/test/generated.py`.  currently they're not added to extra search path, leads `Cannot resolve imported module` error 

this PR tries to improve this

- Breaking change (forces users to change their own code or config): yes, user may observe different lint errors due to generated source also being considered
  generated source files also being considered during linting

### Test plan

- New test cases added

### DIFF
--- a/examples/python/BUILD
+++ b/examples/python/BUILD
@@ -36,3 +36,16 @@ pip_compile(
     requirements_in = "pyproject.toml",
     requirements_txt = "requirements.txt",
 )
+
+
+write_file(
+    name = "root_py_code_generator",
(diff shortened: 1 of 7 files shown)


gregmagolan (dismissed):
see my comment

jbedard (approved):

PR #846, opened 2026-04-26
feat: add clippy in aspect_rules_lint_rust module

dzbarsky:
rules_rs is (in my biased opinion) a more ergonomic and cleaner ruleset that avoids many of the pain points associated with rules_rust usage. It's also a better bzlmod citizen because it doesn't force registration of unwanted toolchains and lead to many unwanted downloads.

Currently rules_rs users who pull in rules_lint get the rules_rust `bazel_dep` and are unable to use the functionality here, while also being forced to register toolchains they don't want or need. I want to be able to recommend rules_lint to my users rather than telling them this ruleset doesn't interop with rules_rs. For example: https://github.com/hermeticbuild/rules_rs/issues/97

## Migration
Existing rules_rust users are encouraged to migrate to rules_rs to realize the benefits, though this module should work for them regardless, on Mac and Linux. On Windows, they will need to configure their `--host_platform` in `.bazelrc`:
- [.bazelrc config](https://github.com/hermeticbuild/rules_rs#bazelrc)
- [platform definitions](https://github.com/hermeticbuild/rules_rs#platformsbuildbazel)


.github/workflows/ci.yaml
```diff
@@ -65,6 +65,7 @@ jobs:
           bazelrc: |
             common     --show_progress_rate_limit=60
             common     --show_timestamps
+            common     --repo_env=ASPECT_TOOLS_TELEMETRY=-all
```
jbedard: why?

dzbarsky: because telemetry is busted and chokes when MODULE.bazel.lock isn't present, which it isn't in CI. So easiest thing is to just disable it. Honestly I am thinking i will just remove telemetry from rules_rs, i got 0 value out of it, only friction...

lint/private/js_linter_inputs.bzl
```diff
@@ -12,20 +12,35 @@ load(
 
 COPY_FILE_TO_BIN_TOOLCHAINS = _COPY_FILE_TO_BIN_TOOLCHAINS
 
+def _bin_file_candidates(target):
```
jbedard: Can we just pin `bazel-lib` within rules_lint for now to workaround this? I'd rather it wasn't part of this PR, sorry my initial attempt to fixing it before this PR seems to have failed :(

examples/cpp/MODULE.bazel
```diff
@@ -12,6 +15,15 @@ local_path_override(
     path = "../..",
 )
 
+rules_lint_go_deps = use_extension("@gazelle//:extensions.bzl", "go_deps", dev_dependency = True)
+rules_lint_go_deps.gazelle_override(
+    build_file_generation = "clean",
+    directives = [
+        "gazelle:proto disable",
```
jbedard: omg... why is this suddenly required? 🤦

dzbarsky: this is what i was mentioning on slack...transitive rules_cc bump bumps protobuf, which is no longer compatible with the version expected in this go module upstream. But we don't even need their protos so just kill them. Gazelle should be configured like this by default...

jbedard: What's less code... this gazelle override or a `single_version_override` for protobuf?

dzbarsky: I think it's about the same. We need an override in each example module just like this... Happy to swap if you prefer

jbedard: But `single_version_override()` on `protobuf` avoids any use of `gazelle`? It would just be:
```
# Explain WTF this is for in one line comment
single_version_override(module_name = "protobuf", version = "X")
```
?

examples/ruby/MODULE.bazel
```diff
@@ -1,14 +1,26 @@
 "Bazel dependencies for Ruby formatting and linting example"
 
 bazel_dep(name = "aspect_rules_lint")
-bazel_dep(name = "rules_ruby", version = "0.21.1")
```
jbedard, line 4: revert

examples/cpp/MODULE.bazel
dzbarsky: not quite that simple, you would first need to add `bazel_dep(protobuf)` and then be able to override. but yeah i suppose it's a few lines shorter. Sounds like you'd prefer I swap, i can try

jbedard: You can pin versions without direct deps on them actually, but it will fail if the dep doesn't exist in your dep graph.

dzbarsky: ah, maybe that's what i was thinking of, thanks for reminding me

lint/rust/MODULE.bazel
```diff
@@ -0,0 +1,33 @@
...
+
+module(
+    name = "aspect_rules_lint_rust",
+    bazel_compatibility = [">=7.6.0"],
+)
+
+bazel_dep(name = "aspect_rules_lint")
+bazel_dep(name = "aspect_rules_js", version = "1.40.0")
+bazel_dep(name = "bazel_lib", version = "3.0.0")
+bazel_dep(name = "llvm", version = "0.7.5")
```
jbedard, line 11: Is this the minimum containing the `archive.org` fix?

lint/rust/MODULE.bazel
```diff
@@ -0,0 +1,33 @@
+"Bazel dependencies for Rust lint support"
+
+module(
+    name = "aspect_rules_lint_rust",
+    bazel_compatibility = [">=7.6.0"],
+)
+
+bazel_dep(name = "aspect_rules_lint")
```
jbedard: Don't we need a minimum `version` here?

lint/rust/clippy.bzl
```diff
@@ -0,0 +1,363 @@
+"""API for declaring a clippy lint aspect that visits rust_{binary|library|test} rules.
+
+Typical usage:
+
+First, install `rules_rs` into your repository, which provisions Clippy integration.
```
jbedard: I thought it worked with `rules_rust` as well and this step wasn't required?

jbedard:
@codex

lint/rust/clippy.bzl
```diff
@@ -0,0 +1,363 @@
...
+)
+use_repo(toolchains, "default_rust_toolchains")
+
+register_toolchains(
+    "@default_rust_toolchains//:all",
+    "@llvm//toolchain:all",
+)
+```
+
+Then set explicit host platforms for operating systems with ABI choices:
```
jbedard: Won't most rules_rs users already have all this?

lint/rust/MODULE.bazel
dzbarsky: we generally just put on latest since we don't make backwards-incompatible changes (and instead fix backwards-incompat changes coming from rules_cc lol). This is whatever version was latest when I first cut this PR :D

lint/rust/MODULE.bazel
dzbarsky: yeah, I think I may have forgotten to save or something...

lint/rust/clippy.bzl
dzbarsky: yep stale docs from original

jbedard (approved):

PR #859, opened 2026-05-08
feat(format): support formatting of go.mod and go.work files

ifutivic:
Adds formatting support for `go.mod` and `go.work` files using [`modfmt`](https://github.com/joshdk/modfmt).

Wiring mirrors aspect-build/rules_lint#795 (Pkl): the language is registered in the central `TOOLS` / `CHECK_FLAGS` / `FIX_FLAGS` tables, file patterns are added to `format.sh`, and a small `examples/go-module/` example shows how to wire up the binary (via `rules_multitool`). No default tool is bundled — users bring their own, same as Pkl, Kotlin, Java, etc.

---

- Breaking change: no

RELEASE NOTES: Supports formatting of `go.mod` and `go.work` files via `modfmt`.

### Test plan

- New test cases added — `format/test/format_test.bats` covers `Go Module` via the mock binary.
- Manual:

  ```
  $ cd examples/go-module
  $ bazel run //tools/format
  $ git diff go.mod
  ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ Linters which are not language-specific:
 | CSS, Less, Sass        | [Prettier]                | [Stylelint]                                             |
 | F#                     | [Fantomas]                |                                                         |
 | Go                     | [gofmt] or [gofumpt]      |                                                         |
+| Go Module              | [modfmt]                  |                                                         |
 | Gherkin                | [prettier-plugin-gherkin] |                                                         |
 | GraphQL                | [Prettier]                |                                                         |
 | HCL (Hashicorp Config) | [terraform] fmt           |                                                         |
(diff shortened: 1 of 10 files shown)


jbedard:
Why `modfmt` instead of https://github.com/shoenig/go-modtool or something else?

ifutivic:
@jbedard I evaluated `go-modtool`, `modfmt` (joshdk), and `modfmt` (abhijit-hota). The one used in this PR is the most complete in my opinion. Both `go-modtool` and the other `modfmt` strip comments, which is a dealbreaker for our monorepo where we occasionally annotate deps that shouldn't be updated, with reasons why.

That said, since there's no standard formatter for `go.mod` files (we even had a custom one at one point), I didn't hardcode a default tool. Users can pick whatever they prefer and override the command name and flags, e.g.:

```starlark
format_multirun(
    name = "format",
    go_module = "//some:go-modtool",
    go_module_check_args = ["--check"],
    go_module_fix_args = ["--fix"],
)
```

I agree it's not the prettiest approach, but keeping `go.mod` consistently formatted is a real pain point for us since all our Go projects share a single file, and without enforcement it quickly devolves into a mess of disjointed `require` blocks. Ideally the standard Go tooling would handle this, but until it does I'd really like to keep this formatter alongside the rest of ours.

jbedard (approved):

PR #862, opened 2026-05-11
fix(format): point format_test failure at the multirun fixer

ifutivic:
A failing `format_test` prints:

> Try running 'bazel run //tools/format:format_test_Go_with_gofmt' to fix this.

…but that's the test target itself, not a fixer. The `FIX_TARGET` env in `_format_attr_factory` is built from the test's own per-language target name (`<format_test_name>_<Lang>_with_<tool>`) instead of the companion `format_multirun`'s fix command.

Add an optional `fix_target` parameter on `format_test` naming the companion `format_multirun` target. When set, `FIX_TARGET` is computed as `<fix_target>_<Lang>_with_<tool>` in that target's package.

Default behavior (parameter omitted) is unchanged.

Usage:

```starlark
format_multirun(name = "format", go = "@aspect_rules_lint//format:gofumpt", ...)
format_test(name = "format_test", fix_target = "format", ...)
```

---

- Breaking change: no

RELEASE NOTES: Properly format error message when formatting fails if the corresponding format target is specified


format/defs.bzl
```diff
@@ -182,8 +190,18 @@ def format_test(name, srcs = None, workspace = None, no_sandbox = False, disable
             srcs = srcs,
         )
 
+    fix_label = native.package_relative_label(fix_target) if fix_target else None
+
     for lang, toolname, tool_label, target_name, custom_args in _tools_loop(name, kwargs):
-        attrs = _format_attr_factory(target_name, lang, toolname, tool_label, "test", disable_git_attribute_checks, custom_args)
+        fix_target_label = None
+        if fix_label:
+            fix_target_label = "//{}:{}_{}_with_{}".format(
```
jbedard: Why do we have to reformat the label when one exists?

Why do we have a random `_with_` in there if the `fix_label` already has a name?

ifutivic: `fix_label` points to a `format_multirun` target that runs multiple formatters in parallel. Since we want to report which specific formatter failed, we construct the name as `format_lang_with_tool`.
We have two targets: `format` and `format_test`. When format_test fails for Go, for example, we want to tell the user to run `format_Go_with_gofmt`, matching what the existing `format_test` failure already reports except with a proper fixer target.

ifutivic: @jbedard Any updates on this? Happy to adjust or rework anything in the PR. My main goal is just making sure the final error message clearly points users toward the correct fix.

ifutivic: @jbedard ping

jbedard: Does that `format_Go_with_gofmt` exist already today? Can we put the `"//{}:{}_{}_with_{}".format(...)` into a util shared somewhere to make that clearer and less fragile?

ifutivic: Yes, it exists today: `format_multirun` creates the per-language fix commands through `_tools_loop`, which names them `<name>_<Lang>_with_<tool>`. So `format_multirun(name = "format", go = ...)` produces a runnable `format_Go_with_gofmt` target.

Good point on the fragility. I pushed a commit that extracts the naming into a `_format_target_name` helper, used both by `_tools_loop` when creating the targets and by `format_test` when computing the fix label. Now there is a single source of truth for the convention, so the suggested target can't drift from the real one.

ifutivic:
The three failing integration tests (kotlin, ruby, starlark) look unrelated to this change. They all fail on the same aspect lint artifact upload step, which needs id-token: write and Aspect credentials that aren't available to fork PRs. The integration-test (format) job and all the format example jobs pass.

jbedard (approved):

ifutivic:
@jbedard I don't have permission to merge the PR so I would be very grateful if you could do it 🙏

jbedard:
Sorry the tests were being flaky and I went to sleep 😅

PR #867, opened 2026-05-15
fix(clang_tidy_wrapper): drop eval, run clang-tidy directly with "$@"

linzhp:
## Summary

  `lint/clang_tidy_wrapper.bash` collapses positional args into a single string and `eval`s it:

  ```bash
  command="$clang_tidy $@ $file > $out_file 2>&1"
  eval $command
  ```

  Once `"$@"` is folded into one string and re-evaluated by the shell, any argument containing shell metacharacters is mis-parsed. In practice this trips on the C/C++
   macro definitions Bazel passes through, e.g.:

  ```
  -DGLOG_DEPRECATED=__attribute__((deprecated))
  -DGLOG_EXPORT=__attribute__((visibility("default")))
  ```

  The `((` is interpreted as bash arithmetic-evaluation syntax, so the wrapper aborts with:

  ```
  clang_tidy_wrapper: eval: line 42: syntax error near unexpected token `('
  ```

  `clang-tidy` then never runs.

  ## Why the lint-only path appears to work

  `clang_tidy_action` uses two execution paths:

  - **`run_shell` (lint mode)** — compiler args are materialized into a Bazel params file (`use_param_file("--config %s", use_always = True)`). The wrapper receives 
  `@params_file`, not the raw flags, so the offending macro definitions never reach the shell.
  - **`run_patcher` (`--fix` / `lint:fix=True` mode)** — compiler args are passed directly on the command line (`args = [..., "--"] + compiler_args`). The full
  `__attribute__((...))` strings hit `eval` and the wrapper crashes.

  End result: lint diagnostics work fine, but `--fix` silently produces an empty patch (`clang-tidy` never executes, so nothing modifies the sandbox source, so the
  diff is empty). The failure is invisible because `JS_BINARY__SILENT_ON_SUCCESS=1` swallows the wrapper's stderr unless the patcher itself exits non-zero, and the
  patcher reports the wrapper's exit code only via `JS_BINARY__EXIT_CODE_OUTPUT_FILE`.

  ## Fix

  Run `clang-tidy` directly with `"$@"` so each argument is preserved verbatim and no shell re-parsing occurs:

  ```bash
  "$clang_tidy" "$@" > "$out_file" 2>&1
  ```

  The `eval` was only there to interpret the `> $out_file 2>&1` redirection inside the assembled string; that's no longer necessary once we run the command directly.

  ## Repro

  A minimal `cc_library` whose `defines` (or transitive `defines`) include an `__attribute__((...))` form is sufficient. With `--@aspect_rules_lint//lint:fix=True
  diff is empty). The failure is invisible because `JS_BINARY__SILENT_ON_SUCCESS=1` swallows the wrapper's stderr unless the patcher itself exits non-zero, and the
  patcher reports the wrapper's exit code only via `JS_BINARY__EXIT_CODE_OUTPUT_FILE`.

  ## Fix

  Run `clang-tidy` directly with `"$@"` so each argument is preserved verbatim and no shell re-parsing occurs:

  ```bash
  "$clang_tidy" "$@" > "$out_file" 2>&1
  ```

  The `eval` was only there to interpret the `> $out_file 2>&1` redirection inside the assembled string; that's no longer necessary once we run the command directly.

  ## Repro

  A minimal `cc_library` whose `defines` (or transitive `defines`) include an `__attribute__((...))` form is sufficient. With `--@aspect_rules_lint//lint:fix=True
  --output_groups=rules_lint_patch`, the `.AspectRulesLintClangTidy.patch` output file will be 0 bytes before this change, and the actual unified diff after.

  ## Test Plan

  - [ ] Lint-only build (`--output_groups=rules_lint_human`) still produces diagnostics.
  - [ ] Fix build (`--output_groups=rules_lint_patch --@aspect_rules_lint//lint:fix=True`) on a target whose compile args contain `__attribute__((...))` produces a
  non-empty patch.
  - [ ] Existing `e2e/clang_tidy` tests pass.
  - [ ] `CLANG_TIDY__VERBOSE=1` still prints the equivalent command line.

---

### DIFF
--- a/lint/clang_tidy_wrapper.bash
+++ b/lint/clang_tidy_wrapper.bash
@@ -33,13 +33,16 @@ else
     out_file=$(mktemp)
 fi
 # include stderr in output file; it contains some of the diagnostics
-command="$clang_tidy $@ $file > $out_file 2>&1"
+# NB: do NOT collapse "$@" into a single string and eval it — args containing
+# shell metacharacters such as `__attribute__((deprecated))` then get re-parsed
+# by the shell ("syntax error near unexpected token `('"). Run clang-tidy
+# directly with "$@" so each argument is preserved verbatim.
 if [[ -n $CLANG_TIDY__VERBOSE ]]; then
     echo "$@"
     echo "cwd: " `pwd`
-    echo $command
+    echo "$clang_tidy $@ > $out_file 2>&1"
 fi
-eval $command
+"$clang_tidy" "$@" > "$out_file" 2>&1
 exit_code=$?
 if [[ -z $CLANG_TIDY__STDOUT_STDERR_OUTPUT_FILE ]]; then
     cat $out_file
--- a/lint/clang_tidy_wrapper.bash
+++ b/lint/clang_tidy_wrapper.bash
@@ -35,13 +35,16 @@ fi
 # Capture raw output here; statistics get filtered into $out_file below.
 raw_out_file=$(mktemp)
 # include stderr in output file; it contains some of the diagnostics
-command="$clang_tidy $@ $file > $raw_out_file 2>&1"
+# NB: do NOT collapse "$@" into a single string and eval it — args containing
+# shell metacharacters such as `__attribute__((deprecated))` then get re-parsed
+# by the shell ("syntax error near unexpected token `('"). Run clang-tidy
+# directly with "$@" so each argument is preserved verbatim.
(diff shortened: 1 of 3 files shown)


jbedard:
Any way to reproduce it in a test/example?

Please rebase

linzhp:
Added a unit test and rebased. PTAL

jbedard (approved):

PR #875, opened 2026-05-21
fix(ty): resolve imports from standard dependencies without imports attribute

mp4096:
### Problem

When using the `ty` aspect to lint Python targets that depend on standard external Bazel dependencies (like `abseil-py` or `rules_python` pip dependencies that don't set the `imports` attribute), `ty` fails with `unresolved-import` errors for those dependencies.

This happens because:
1. Standard Bazel dependencies package their code directly at their repository root (e.g., `absl` is a folder directly under the repository root) and do not set the `imports` attribute in their `py_library` rules, relying on Bazel's default behavior instead.
2. Consequently, `dep[PyInfo].imports` is empty.
3. The `ty` aspect implementation only populated `extra-search-path` using `dep[PyInfo].imports`. Since it was empty, the dependency's repository root (e.g., `external/abseil-py+`) was omitted from the linter's search paths.

### Solution

This PR introduces a fallback to `dep.label.workspace_root` (e.g., `external/abseil-py+`) when `dep[PyInfo].imports` is empty. 
This ensures the dependency's workspace root (and its `bazel-bin` counterpart for generated files) is correctly added to `ty`'s `--extra-search-path` arguments.

### Test Plan

Added a regression test in `examples/python`:
1. Added `abseil-py` to `examples/python/MODULE.bazel`.
2. Created `examples/python/test/dep_bcr.py` importing `absl.testing`.
3. Added `ty_should_find_bcr_dep` test target in `examples/python/test/BUILD`.

Before this fix, running this test fails with:
```
error[unresolved-import]: Cannot resolve imported module `absl.testing`
```

With this fix, the test passes successfully.

### DIFF
--- a/examples/python/test/BUILD
+++ b/examples/python/test/BUILD
@@ -113,41 +113,44 @@ pydoclint_test(
     expected_exit_code = 1,
 )
 
+py_library(
+    name = "generated",
+    srcs = ["generated.py"],
+    imports = ["."],
+)
 
-py_library(name = "generated",
-           srcs = ["generated.py"],
-           imports = ["."])
-
-py_library(name = "dep_generated",
-           srcs = ["dep_generated.py"],
-           deps = [":generated"])
+py_library(
+    name = "dep_generated",
+    srcs = ["dep_generated.py"],
+    deps = [":generated"],
+)
 
 ty_test(
     name = "ty_should_find_generated_dep",
-    srcs = [":dep_generated"]
+    srcs = [":dep_generated"],
 )
 
-py_library(name = "dep_ext_generated",
-           srcs = ["dep_generated.py"],
-           deps = ["@ext_mod//src:generated"]
+py_library(
+    name = "dep_ext_generated",
+    srcs = ["dep_generated.py"],
+    deps = ["@ext_mod//src:generated"],
 )
 
 ty_test(
     name = "ty_should_find_ext_generated_dep",
-    srcs = [":dep_ext_generated"]
+    srcs = [":dep_ext_generated"],
 )
 
-py_library(name = "dep_bcr",
-           srcs = ["dep_bcr.py"],
-           deps = ["@abseil-py//absl/testing:absltest"]
+py_library(
+    name = "dep_bcr",
+    srcs = ["dep_bcr.py"],
+    deps = ["@abseil-py//absl/testing:absltest"],
 )
 
 ty_test(
     name = "ty_should_find_bcr_dep",
-    srcs = [":dep_bcr"]
+    srcs = [":dep_bcr"],
 )
-
-
 
 ty_test(
     name = "ty_should_fail_unsupported_operator",
--- a/examples/python/test/BUILD
+++ b/examples/python/test/BUILD
@@ -137,6 +137,18 @@ ty_test(
     srcs = [":dep_ext_generated"]
 )
 
+py_library(name = "dep_bcr",
+           srcs = ["dep_bcr.py"],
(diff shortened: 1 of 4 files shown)


examples/python/test/BUILD
```diff
@@ -137,6 +137,18 @@ ty_test(
     srcs = [":dep_ext_generated"]
 )
 
+py_library(name = "dep_bcr",
```
udaya2899: please fix weird formatting here

mp4096: @jbedard @JesseTatasciore the `py_library` formatting in this `BUILD` file is consistently weird, should i fix it while at it or does it have a deeper meaning?

udaya2899:
Friendly pinging @jbedard / @JesseTatasciore  to look into this. Integrating `ty` via `rules_lint` has been really good for us, except for this bug. Merging this would help us to use this widely across the codebase. Please let me know if you still need anything from us.

faximan:
I'd also really like to see this bug fixed!

examples/python/test/BUILD
jbedard: I have no idea. If CI doesn't fail then follow the consistently weird conventions like you have.

Really `buildifier` should automate the formatting and we shouldn't have to think about it

lint/ty.bzl
```diff
@@ -180,10 +180,15 @@ def _ty_aspect_impl(target, ctx):
...
-                    resolved = _resolve_import_path(import_path, ctx.workspace_name, ctx.bin_dir.path)
-                    for e in resolved:
-                        import_paths[e] = True
+                imports = dep[PyInfo].imports.to_list()
+                if imports:
+                    for import_path in imports:
+                        resolved = _resolve_import_path(import_path, ctx.workspace_name, ctx.bin_dir.path)
+                        for e in resolved:
+                            import_paths[e] = True
+                elif dep.label.workspace_root:
```
jbedard, line 189: Why do we only do this if `imports` is falsy?

udaya2899: #### Example Scenario:

  Imagine an external dependency  @my_lib  with this filesystem layout:

    external/my_lib/
    ├── BUILD.bazel
    └── src/
        └── my_lib/
            ├── __init__.py
            └── core.py

  And its  BUILD.bazel  declares:

    py_library(
        name = "my_lib",
        srcs = glob(["src/**/*.py"]),
        imports = ["src"], # <--- Explicitly overrides import path
    )

  Because of  imports = ["src"] , the correct way to import this library is:

    import my_lib

and NOT `import src.my_lib`

  #### If we used  if instead of elif  (Unconditionally adding the root  external/my_lib ):

  1. Type-Checking (Lint) Time: The linter's search path would include both  external/my_lib/src  (from  imports) AND  external/my_lib  (from  workspace_root ).
  2. A developer could write:
    import src.my_lib # This is WRONG, but the linter would mark it as VALID!

linter would pass, but Bazel run would fail

mp4096: @udaya2899 thanks, i was just about to reply myself.

@jbedard yes indeed it works with always adding

```py
import_paths[dep.label.workspace_root] = True
import_paths[ctx.bin_dir.path + "/" + dep.label.workspace_root] = True
``` 

unconditionally, but i guess it's better to keep the paths tight.

examples/python/test/BUILD
mp4096: I ran `buildifier` on the file.

jbedard (approved):

PR #881, opened 2026-05-27
chore: bump aspect-cli to v2026.22.17

gregmagolan:
Bump aspect-cli to v2026.22.17. Also drops the now-unnecessary `--bazel-flag=--output_groups=rules_lint_human` flag from `aspect lint` invocations in the e2e bats tests, and fixes a latent SARIF conversion bug exposed by that change.

**Background:** the old explicit `--output_groups=rules_lint_human` suppressed the machine-readable output path entirely, so `parse_to_sarif_action` never ran for Ruby targets. The newer `aspect lint` requests `rules_lint_machine` (the SARIF-producing group) by default, which invoked the SARIF converter — and found that `AspectRulesLintRuboCop` and `AspectRulesLintStandardRB` were missing from the switch in `tools/sarif/sarif.go` (both linters were added without registering their format). Adds a JSON parser for the RuboCop `--format json` output to fix the conversion.

See https://github.com/aspect-build/aspect-cli/releases/tag/v2026.22.17

---

### Test plan

- CI


examples/cpp/test/lint_test.bats
```diff
@@ -9,7 +9,7 @@ function assert_cpp_lints() {
 }
 
 @test "should produce reports" {
-	run aspect lint --bazel-flag=--config=lint --bazel-flag=--output_groups=rules_lint_human -- //src/...
```
jbedard, line 12: Isn't this required?

gregmagolan: It was only required because of a bug in the lint task. With the bump in aspect cli version --output_groups=rules_lint_{human|machiine} is always set now.

Issue #538, opened 2025-05-10
[FR]: Support biome as linter (and probably formatter)


daivinhtran:
### What is the current behavior?

We don't have a rule for [biome](biomejs.dev) yet.

### Describe the feature

[biome](biomejs.dev) is becoming the faster alternative to eslint and prettier. I'd be great if `rules_lint` provides a support for biome.

daivinhtran:
@arrdem I have some time to implement and ship this now. Feel free to assign the issue to me.

PR #887, opened 2026-05-30
feat: Add biome linter/formatter

SiarheiFedartsou:
---

### Changes 

Issue: https://github.com/aspect-build/rules_lint/issues/538

- added Biome as formatter (Prettier is still default, but it can be overriden) 
- added Biome as linter
- extended tests

I was mainly following ESLint/Prettier as an example.

NOTE: I am very new to this project, so I was heavily relying on AI to help me with this PR, I hope it is acceptable.

### Test plan

New tests added.

### DIFF
--- a/tools/sarif/sarif.go
+++ b/tools/sarif/sarif.go
@@ -38,7 +38,7 @@ func ToSarifJsonString(label string, mnemonic string, report string) (sarifJsonS
 	regex := regexp.MustCompile(`^{\s+"\$schema":.+sarif`)
 	// If it's already in SARIF format, just return it
 	if regex.Match([]byte(report)) {
-		return report, nil
+		return normalizeSarifUris(label, report)
 	}
 
 	if len(mnemonic) == 0 {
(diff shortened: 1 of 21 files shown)


tools/sarif/sarif.go
```diff
@@ -36,9 +36,9 @@ func mnemonicPrettyName(mnemonic string) string {
 
 func ToSarifJsonString(label string, mnemonic string, report string) (sarifJsonString string, err error) {
 	regex := regexp.MustCompile(`^{\s+"\$schema":.+sarif`)
-	// If it's already in SARIF format, just return it
+	// If it's already in SARIF format, normalize paths before returning it.
 	if regex.Match([]byte(report)) {
-		return report, nil
+		return normalizeSarifUris(label, report)
```
SiarheiFedartsou, line 41: It seems it always was the issue that paths there are not normalized...

jbedard: I don't understand this. Previously we were simply returning the `report` when it matched a regex, now we are doing `json.Unmarhsal` on that report if it matches that same regex? Something seems wrong here...

tools/sarif/sarif.go
```diff
@@ -257,6 +289,16 @@ func determineRelativePath(path string, label string) string {
 	if bazel_package != "" {
 		re = regexp.MustCompile(`\/execroot\/[^\/]+\/(` + bazel_package + `\/.*)$`)
 	}
+
+	binRe := regexp.MustCompile(`\/execroot\/[^\/]+\/bazel-out\/[^\/]+\/bin\/(.*)$`)
```
jbedard, line 293: Why would biom be the first tool to run into this scenario? Can you do some debugging to determine which paths are being outputted incorrectly here?

Wild guess: are you on a mac and biome is escaping the sandbox so you get odd paths? In which case I wouldn't want to support this and would rather wait for proper macos sandboxing (coming soon! 🤞)

PR #900, opened 2026-06-08
fix(clang-tidy): strip summary statistics so a clean run yields an empty report

### LINKED ISSUE #899: [Bug]: clang-tidy lint_test fails under fail_on_violation when all findings are suppressed
### What happened?

A `clang_tidy` lint test on a **clean** target fails under `--@aspect_rules_lint//lint:fail_on_violation` even though clang-tidy finds no real violations and exits 0. The entire test output is clang-tidy's benign summary:

```
FAIL: //modules/cpp_library:cpp_library.lint
31344 warnings generated.
Suppressed 31344 warnings (31341 in non-user code, 3 with check filters).
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
```

Every warning was suppressed (system / non-user headers plus `.clang-tidy` check filters); there are zero user-facing diagnostics and clang-tidy's exit code is 0. The same target **passes** when `fail_on_violation` is not set.

### Version

```
Linux x86_64
```

Output of `bazel --version`:

```
bazel 8.7.0
```

Version of the Aspect rules:

```starlark
bazel_dep(name = "aspect_rules_lint", version = "2.5.2")
```

`main`'s `lint/clang_tidy_wrapper.bash` is also affected — it neither filters these lines nor passes `--quiet`.

```
C++ (clang-tidy 21.1.8)
```

### How to reproduce

```shell
# Any clang_tidy aspect over sources that include system headers, with a
# .clang-tidy that suppresses checks, run with fail_on_violation:
bazel test --test_tag_filters=lint --@aspect_rules_lint//lint:fail_on_violation //...
```

### Any other information?

- Under `fail_on_violation`, `output_files()` (`lint/private/lint_aspect.bzl`) emits no `.exit_code` file, so `lint/lint_test.sh` uses `assert_output_empty` — i.e. *any non-empty `.out` report counts as a violation*.
- `lint/clang_tidy_wrapper.bash` captures stdout **and** stderr (`> $out_file 2>&1`). clang-tidy always prints its statistics to **stderr** — `N warnings generated.`, `Suppressed N warnings (…)`, and the `Use -header-filter=…` hint — even when it exits 0 with nothing user-facing.
- Result: the report is non-empty, so `assert_output_empty` fails. This breaks the framework's "empty report ⟺ clean" contract that ruff/eslint/etc. honor.

I verified on clang-tidy 21.1.8 that real diagnostics go to **stdout** while these statistics go to **stderr**, so the noise can be filtered without losing diagnostics (including the `clang-diagnostic-error` lines the wrapper's own fatal-error scan relies on).

Funnily enough, the SARIF converter already discards these lines (errorformat `%-G%m` plus the `Filename != "" && Text != ""` writer guard in `tools/sarif/sarif.go`; the golden in `sarif_test.go` expects 2 results from an input that contains the summary noise). Only the human report contains it, because the wrapper hands raw stdout+stderr straight to `assert_output_empty` with no parser in between.

Happy to send a PR: strip the non-diagnostic summary/hint lines in `clang_tidy_wrapper.bash` so a suppressed-only run yields an empty report, preserving diagnostics and fatal-error detection. (`--quiet` alone is insufficient — it does not suppress the frontend's `N warnings generated.` line.)

vadikmironov:
## Summary

clang-tidy prints summary statistics — `N warnings generated.`, `Suppressed N warnings (...)`, the `Use -header-filter=...` hint — even when every finding is suppressed and it exits 0. `clang_tidy_wrapper.bash` captured these into the human report, so under `--@aspect_rules_lint//lint:fail_on_violation` (which treats a non-empty report as a violation) a perfectly clean target fails. Default mode passes because it keys off clang-tidy's exit code rather than report emptiness.

Fixes #899.

## Change

- Filter the non-diagnostic summary/hint lines in `clang_tidy_wrapper.bash`. Real diagnostics — including the `clang-diagnostic-error` lines the wrapper's own fatal-error detection relies on — are preserved.
- Add an integration regression test to `examples/cpp/test/lint_test.bats`: it runs the real clang-tidy aspect over `//src:hello_cc` and asserts the report keeps real diagnostics but excludes the summary statistics (`warnings generated`, `Suppressed`). Confirmed it fails without the wrapper change.

## Notes

- `--quiet` alone is insufficient: it drops clang-tidy's `Suppressed ...` stats and the hint, but not the Clang frontend's `N warnings generated.` line.
- This brings the human report in line with the machine/SARIF path, which already discards these lines via the errorformat `%-G%m` rule plus the `Filename != "" && Text != ""` writer guard in `tools/sarif/sarif.go`.


lint/clang_tidy_wrapper.bash
```diff
@@ -32,15 +32,22 @@ if [[ -n $CLANG_TIDY__STDOUT_STDERR_OUTPUT_FILE ]]; then
...
     echo "$@"
     echo "cwd: " `pwd`
     echo $command
 fi
 eval $command
 exit_code=$?
+# Drop clang-tidy summary statistics (e.g. "N warnings generated.") that it
+# prints even on a clean, exit-0 run thus tripping in fail_on_violation mode.
+# Diagnostics are kept.
+grep -Ev '^[0-9]+ (warnings?|errors?)( and [0-9]+ errors?)? generated\.$|^Suppressed [0-9]+ warnings? \(.*\)\.$|^Use -header-filter=.*$|^[0-9]+ warnings? treated as errors$' $raw_out_file > $out_file || true
```
jbedard: What is the 'Use -header-filter' one?

vadikmironov: It's the usage hint clang-tidy prints immediately after the Suppressed N warnings line whenever it has suppressed anything. Here is an example from my other repo:

```
Executing tests from //modules/cpp_app:cpp_app.lint
-----------------------------------------------------------------------------
103762 warnings generated.
Suppressed 103762 warnings (103759 in non-user code, 3 with check filters).
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
```

Like the other two patterns, it's a statistics/usage line rather than a diagnostic. clang-tidy emits it to stderr as part of the suppression summary (in this PR's test fixture it accompanies Suppressed 46749 warnings (46749 in non-user code)). Since the wrapper folds stderr into the report, this line is part of what makes an otherwise-clean run's report non-empty and trips fail_on_violation.

jbedard: Dropping an one of these `|`-separated expressions cause the test to fail?

vadikmironov: Fair enough, the test only asserted two of the branches. Added a `Use -header-filter` check to the bats test + a dedicated warnings-as-errors test asserting the `error:` diagnostic survives while the summary line is stripped + tweaked grep for matching singular 'error' word. Any of the grep branches missing now would trigger a fail.

lint/clang_tidy_wrapper.bash
```diff
@@ -32,15 +32,22 @@ if [[ -n $CLANG_TIDY__STDOUT_STDERR_OUTPUT_FILE ]]; then
...
     echo "$@"
     echo "cwd: " `pwd`
     echo $command
 fi
 eval $command
 exit_code=$?
+# Drop clang-tidy summary statistics (e.g. "N warnings generated.") that it
+# prints even on a clean, exit-0 run thus tripping in fail_on_violation mode.
+# Diagnostics are kept.
+grep -Ev '^[0-9]+ (warnings?|errors?)( and [0-9]+ errors?)? generated\.$|^Suppressed [0-9]+ warnings? \(.*\)\.$|^Use -header-filter=.*$|^[0-9]+ warnings? treated as errors?$' $raw_out_file > $out_file || true
```
jbedard: I think the `|| true` swallows real errors (grep exit code `2`)?

A robot told me `|| [ $? -eq 1 ]` is better to allow `0|1` while still failing on `2`...?

vadikmironov: Your robot is technically right, but possibly missing that there is no set -e/pipefail in this script and whatever non-zero return from grep might be it will be ignored. Which actually is a real issue for case of retcode == 2 which must be handled. There is an exit_code var being set and checked later from the actual clang_tidy and what I can do either try and weave grep return code into it or just straighforward

```
grep_status=$?
if [ "$grep_status" -gt 1 ]; then
```
right after grep command. Do you have any preference?

vadikmironov: Went with the straightforward grep_status guard, pushed. Please let me know if this looks better.

jbedard (approved):

vadikmironov:
@jbedard , thanks a lot for your comments and merging this!

PR #901, opened 2026-06-08
feat(go): hermetic golangci-lint linter aspect

b3cramer:
## Summary

Adds **golangci-lint** as a first-class rules_lint linter aspect for Go — the first Go linter in the ruleset (previously Go had only formatters). It runs the real golangci-lint binary as a hermetic, per-package, remote-cache-friendly Bazel action.

The hard part: golangci-lint is type-aware and loads packages through `go/packages`, which normally shells out to `go list`/`go build` — impossible in a hermetic sandbox. rules_go's stock `gopackagesdriver` doesn't help (it shells out to `bazel query`/`bazel build` at runtime). The solution here is a **static, file-reading `GOPACKAGESDRIVER`**: a rules_lint aspect reuses rules_go's `go_pkg_info_aspect` to pre-compute each package's metadata as declared action inputs, and a vendored fork of rules_go's driver *read-half* feeds golangci-lint from those staged files — no bazel-in-bazel, no network, no shared module cache.

### What's included
- **Static driver** (`lint/private/gopackagesdriver_static/`): vendored rules_go read-half (verbatim, with provenance headers) + a new `main.go` that injects stdlib `ExportFile` from precompiled `.a` archives, takes roots from `GOPACKAGESDRIVER_ROOTS`, and resolves `go_deps` imports by `PkgPath`. One local modification to `packageregistry.go` (clearly marked) for go_deps with empty `Imports` maps.
- **The aspect** (`lint/golangci_lint.bzl`, `lint_golangci_lint_aspect`): visits `go_library`/`go_binary`/`go_test`; lints only the target package and consumes all dependencies (stdlib + external) as **export data only**; normalizes golangci-lint's SARIF URIs to workspace-relative paths.
- **Provisioning**: golangci-lint v2.12.2 via `@multitool`, exposed as `//lint:golangci_lint_bin`.
- **Example + tests** (`examples/go/`): wires the aspect via `--config=lint`; `lint_test`s assert a real `govet` finding on `//src:hello` (exit 1) and a clean case on `//src:greeting` (exit 0).
- **Docs**: README supported-tools row.

### Design decisions worth a reviewer's eye
- **Export-only dependencies** — correct linter semantics (deps are compiled artifacts, not lint targets) and avoids false typecheck errors from dependency source that imports outside the Bazel graph.
- **`GOMAXPROCS=1`** mitigates a known x/tools gcimporter data race when decoding export data; documented as a pragmatic mitigation, not a hard guarantee.
- **Per-package action granularity** gives incremental, remote-cacheable linting — editing one package does not invalidate another's lint result (verified).
- **Deferred**: golangci-lint `--fix` patch mode (mutates in place rather than emitting a diff).

## Test Plan
- [x] `bazel test //lint/private/gopackagesdriver_static:gopackagesdriver_static_test` (driver protocol unit test)
- [x] `cd examples/go && bazel test //test:all` — `golangci_lint_finds_issue` (exit 1) + `golangci_lint_clean` (exit 0)
- [x] Per-package cache reuse verified (changing one package leaves another's report byte-identical)
- [ ] CI Linux (x86_64/arm64) — only verified locally on macOS arm64; the multitool lockfile includes Linux binaries but the Linux execution path is exercised by CI


lint/golangci_lint.bzl
```diff
@@ -0,0 +1,356 @@
...
+        progress_message = "Normalizing golangci-lint SARIF URIs for %{label}",
+    )
+
+# A small python3 post-processor. Each result URI is rewritten to the path
+# following "/execroot/<repo>/" (the portable, workspace-relative path). A URI
+# with no "/execroot/" segment (e.g. non-sandboxed local execution that already
+# emitted a relative path) is returned as-is apart from stripping a leading
+# "./". Empty input (no findings / empty file) yields a minimal valid empty
+# SARIF document.
+_SARIF_NORMALIZE_SCRIPT = """\
```
rafikk, line 214: Can this be done in `//tools/sarif` to avoid a dependency on a non-hermetic system python?

lint/private/gopackagesdriver_static/main.go
```diff
@@ -0,0 +1,197 @@
...
+	"strings"
+
+	"golang.org/x/tools/go/packages"
+)
+
+// RulesGoStdlibLabel is referenced by PackageRegistry.Match. The bazel half of
+// the upstream driver defined it; the static driver never receives this label
+// as a root (stdlib packages are pulled in transitively by the import walk),
+// but the symbol must exist for the vendored registry to compile.
+const RulesGoStdlibLabel = "@io_bazel_rules_go//:stdlib"
```
jbedard, line 39: Why are we hardcoding repo names in here? :/

PR #908, opened 2026-06-15
refactor(lint): accept args objects in run_patcher

rafikk:
This PR amends the implementation of `run_patcher` to accept an Args object created via `ctx.action.args()`. The motivation for this is that linting actions often have divergent implementations for the patch vs. check path where command-line argument construction cannot be shared. Now common arguments can be shared between the two and we get the other benefits of using Args objects (e.g. deferred evaluation of depsets to execution phase, etc.).

The change is to spill out the args object into a params file in `run_patcher` and load the command line arguments from a separate file rather than from the patcher config JSON object. Passing a list continues to work and the behavior should be unchanged. I've also updated a handful of linter aspects to pass Args objects instead.

### Changes are visible to end-users:

No, this is an internal API refactoring

### Test plan

### DIFF
--- a/lint/buildifier.bzl
+++ b/lint/buildifier.bzl
@@ -49,7 +49,10 @@ def buildifier_action(ctx, executable, srcs, stdout = None, exit_code = None, pa
     """
     if patch != None:
         wrapper = ctx.actions.declare_file(ctx.label.name + ".buildifier_wrapper.sh")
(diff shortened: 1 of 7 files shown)


jbedard (approved):
👍 even just for dropping the `.path` usage for potential future path-mapping support

PR #909, opened 2026-06-16
fix(release): fail release if any sarif_parser platform integrity is missing

gregmagolan:
## Summary

2.7.0 shipped with an incomplete `tools/integrity.bzl` (missing `sarif_parser-linux_amd64`), which broke `aspect lint`/`aspect format` on Linux for consumers:

```
no such package '@@aspect_rules_lint++toolchains+sarif_parser_linux_amd64//':
key "sarif_parser-linux_amd64" not found in dictionary
```

### Same root cause as rules_py #1110

Yes — this is the **same issue** @jbedard fixed in aspect-build/rules_py@9633720ec85fe97d2a3300b346b6101a6cdb64d8 (#1110). GHA `download-artifact` now extracts a single artifact **flat** into the workspace root rather than into `<name>/` (actions/download-artifact#455). rules_lint uses the same `release_ruleset.yaml@v7.6.0` reusable workflow and reads `go-binaries/*.sha256`, so after the change that glob matched nothing → empty `RELEASED_BINARY_INTEGRITY` → the broken release.

## What changed

1. **Root-cause fix (ports rules_py #1110):** relocate the flat-extracted `*.sha256` (and their binaries) back into `go-binaries/` before the integrity glob runs, so `release_prep.sh` and `release_files: go-binaries/*` find them again.

2. **Guardrail (new):** after generating `tools/integrity.bzl`, fail the release if any platform declared in `SARIF_PARSER_PLATFORMS` is missing its integrity key. This catches *any* future cause of a partial dict at publish time instead of shipping it broken.

## Are these changes visible to end-users?

No — release tooling. (It prevents a class of broken release that is very visible.)

## Test plan

- Verified the validation against fixtures: an incomplete dict (the broken 2.7.0 shape) fails listing every missing platform; a complete dict (all 7 `SARIF_PARSER_PLATFORMS`) passes.
- Confirmed rules_lint uses the same `release_ruleset.yaml@v7.6.0` as rules_py, so it has the same flat-extract behavior #1110 addressed.
- `shellcheck` clean (no new error-level findings).

> [!NOTE]
> A 2.7.1 patch release with the corrected integrity dict is still needed to unbreak existing 2.7.0 consumers; this PR prevents recurrence.

### DIFF
--- a/.github/workflows/release_prep.sh
+++ b/.github/workflows/release_prep.sh
@@ -40,6 +40,30 @@ RELEASED_BINARY_INTEGRITY = $(
 )
 EOF
 
+# Fail the release if any sarif_parser platform the toolchain requires is missing
+# an integrity entry. Without this, a dropped or unbuilt release binary silently
+# produces an incomplete RELEASED_BINARY_INTEGRITY, and the gap only surfaces at
+# `bazel build` time in a consuming repo as
+# "key \"sarif_parser-<platform>\" not found in dictionary" (see #906).
+# The required platforms are the keys of SARIF_PARSER_PLATFORMS.
+missing=()
+for platform in $(grep -oE '^    "[a-z0-9_]+": struct\(' tools/toolchains/sarif_parser_toolchain.bzl | sed -E 's/^    "([^"]+)".*/\1/'); do
+  ext=""
+  [[ "$platform" == windows_* ]] && ext=".exe"
+  key="sarif_parser-${platform}${ext}"
+  if ! grep -q "\"${key}\"" ${PREFIX}/tools/integrity.bzl; then
+    missing+=("$key")
+  fi
+done
+if [[ ${#missing[@]} -gt 0 ]]; then
+  echo "ERROR: release integrity is missing entries for: ${missing[*]}" >&2
+  echo "The go-binaries/ artifact did not contain a .sha256 for every required" >&2
+  echo "sarif_parser platform. Aborting to avoid publishing a broken release." >&2
+  echo "Generated tools/integrity.bzl was:" >&2
+  cat ${PREFIX}/tools/integrity.bzl >&2
+  exit 1
+fi
+
 # Append that generated file back into the archive
 tar --file $ARCHIVE_TMP --append ${PREFIX}/tools/integrity.bzl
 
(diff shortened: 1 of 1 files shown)


jbedard:
Is this the same issue as https://github.com/aspect-build/rules_py/commit/9633720ec85fe97d2a3300b346b6101a6cdb64d8?

gregmagolan:
> Is this the same issue as [aspect-build/rules_py@9633720](https://github.com/aspect-build/rules_py/commit/9633720ec85fe97d2a3300b346b6101a6cdb64d8)?

Same symptom family, different root cause — and different fix. Both are about release_prep.sh generating an incomplete integrity dict, which then fails downstream with a "missing key/platform" error at consumer build time. But the why differs:

rules_py 9633720 (Jason's): The cause was a GHA download-artifact behavior change (actions/download-artifact#455) — it now extracts a single artifact flat into the workspace root instead of artifacts-<name>/. So the *.sha256 files weren't where the glob expected them → empty/partial dict. His fix relocates the flattened files so the existing glob finds them.

rules_lint #909 (mine): I addressed it from the other side — release_prep.sh has no validation that all required platforms ended up in the dict, so any cause (the download-artifact flattening, a failed binary build, etc.) silently produces a broken release. My fix asserts every SARIF_PARSER_PLATFORMS entry is present and fails the release otherwise.

So: the rules_lint 2.7.0 breakage was very likely caused by the exact same download-artifact change Jason fixed in rules_py. My #909 is complementary — it's the guardrail that would have caught it (and will catch the next cause), whereas Jason's is the specific root-cause fix. The ideal rules_lint fix is both: port Jason's download-artifact relocation and keep the validation.

.github/workflows/release_prep.sh
```diff
@@ -28,6 +28,17 @@ tar --create --auto-compress \
 # Delete the placeholder file
 tar --file $ARCHIVE_TMP --delete ${PREFIX}/tools/integrity.bzl
 
+# download-artifact now extracts a single artifact flat into the workspace root
```
jbedard: Edit the comments to be useful to someone in the future, and not sound like a dumb robot. "now" is quite meaningless when someone reads this in 1y. Either delete such silly comments or edit them to not be useless in 1y

jbedard (approved):

PR #911, opened 2026-06-16
fix(keep-sorted): honor 'no-lint' tag

rafikk:
This PR updates `lint_keep_sorted_aspect` to honor the `no-lint` tag on targets.

Since the keep-sorted aspect runs over all rule kinds, it did not include a `should_visit` check in its implementation, which causes it to ignore `no-lint` tags on the target.

I've updated `should_visit` to interpret `["*"]` to mean "allow all rule kinds"

- Searched for relevant documentation and updated as needed: no, it's a bug fix

### Test plan

- Manual testing; please provide instructions so we can reproduce: Tested against our internal repository. Let me know if you'd like me to add a test case for this.

### DIFF
--- a/lint/keep_sorted.bzl
+++ b/lint/keep_sorted.bzl
@@ -28,7 +28,7 @@ following the documentation at https://github.com/google/keep-sorted#usage.
 """
 
 load("@jq.bzl//jq:jq.bzl", "jq_lib")
-load("//lint/private:lint_aspect.bzl", "LintOptionsInfo", "OPTIONAL_SARIF_PARSER_TOOLCHAIN", "OUTFILE_FORMAT", "filter_srcs", "noop_lint_action", "output_files", "parse_to_sarif_action", "patch_and_output_files")
+load("//lint/private:lint_aspect.bzl", "LintOptionsInfo", "OPTIONAL_SARIF_PARSER_TOOLCHAIN", "OUTFILE_FORMAT", "filter_srcs", "noop_lint_action", "output_files", "parse_to_sarif_action", "patch_and_output_files", "should_visit")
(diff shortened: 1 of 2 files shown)


jbedard:
Can you add a test reproducing the bug and showing it's fixed, most likely just a single target somewhere in `examples/`?

rafikk:
> Can you add a test reproducing the bug and showing it's fixed, most likely just a single target somewhere in `examples/`?

I can add a test, but I just want to note that no other linters have a test for the no-lint tag :)

It's also a bit tricky with the current testing setup where, for example, `machine_report_rule` requires the aspect output group, which doesn't exist for `no-lint` targets, turning the tests into a build error.

jbedard:
This scenario seemed more complicated then regular `no-lint` tag scenarios though? If a test is a hassle we can skip it, it just means it will break without notice in the future 😅

rafikk:
> This scenario seemed more complicated then regular `no-lint` tag scenarios though?

No, it's exactly the same. The aspect was just missing the `should_visit` call, which all other linter aspects have.

The main functional change here is to treat `["*"]` in `should_visit` as "visit all rules", but that's already covered by existing test cases. For example, initially, I had `should_visit` treat `[]` as "visit all rules", but the integration tests for QML caught a regression with that change.

In terms of test coverage, `no-lint` appears [only twice](https://github.com/search?q=repo%3Aaspect-build%2Frules_lint%20no-lint&type=code) in this repo: in the implementation of `should_visit` and in the documentation, so there are no additional unit or integration tests for that behavior.

IMO, this change is ready to merge without an additional test :)

jbedard (approved):

jbedard:
> IMO, this change is ready to merge without an additional test :)

Only because you promised you'd expand the coverage later 😉

PR #914, opened 2026-06-17
feat(lint): add scalafix linter

rafikk:
Supports Scalafix as a linter for Scala code. Scalafix is a refactoring and linting tool from Scala Center that supports both syntactic and semantic code analysis.

Scalafix produces lint diagnostics and rewrite patches for configured rules.

This aspect supports the built-in syntactic and semantic rules. We can amend it in the future to support custom plugins, but I have not yet had a need for this.

Notes:
* Scalafix only produces output in one mode which consists of a mixture of diagnostic lines and rewrite patches. When a configured Scalafix rule is able to produce rewrites, it prints a patch but _does not_ print a diagnostic line, meaning the SARIF reports that this aspects produce are not as precise as would be ideal. We can only produce a generic "Scalafix rewrite suggested" line for diagnostics with rewrites. I've opened an upstream issue to discuss producing diagnostic-only ouput and hope to find time to work on it: https://github.com/scalacenter/scalafix/issues/2459
* This PR is based on draft in #711 which stalled. I've used the original PR as a starting point.

#### Suggested Release Notes

Scala targets can not be linted using scalafix.

### Test plan

- New test cases added


MODULE.bazel
```diff
@@ -34,6 +34,7 @@ bazel_dep(name = "rules_multitool", version = "0.11.0")
 bazel_dep(name = "rules_shell", version = "0.5.0")
 bazel_dep(name = "rules_java", version = "7.0.6")
 bazel_dep(name = "rules_cc", version = "0.2.0")
+bazel_dep(name = "rules_scala", version = "7.1.6")
```
jbedard: I don't want to add any new deps here as this now effects the module versions of any rules_lint user such as [pulling in rules_java](https://github.com/bazel-contrib/rules_scala/blob/30a76711766f11a200e7c5c3e79a4a788e1b84cc/MODULE.bazel#L52) - who cares if the rules_lint user doesn't use it, but a rules_lint user might not want rules_java being bumped due to a linter.

I'm proposing any lint tools that require a `bazel_dep` (other then maybe skylib or bazel-lib) live in their own MODULE. See the [new aspect_rules_lint_rust MODULE](https://github.com/aspect-build/rules_lint/tree/main/lint/rust) for the first example - that one is now published to BCR and everything.

In rules_lint v3 I'm proposing everything move out of this root MODULE, for now I just want to avoid adding new ones.

rafikk: Thanks for the review @jbedard.

I've moved the scalafix linter into a new module `aspect_rules_lint_scala` and updated the example code.

Note that this PR still requires changes to the SARIF processor in `rules_lint`, so the flow here has to be: publish a new `rules_lint` release with the changes, bump the minimum rules_lint version for scala, and then publish the new `aspect_rules_lint_scala` release. Let me know if that sounds right to you.

rafikk:
The CI failure for [CI - Vanilla Bazel / integration-test (examples/scala) (pull_request)](https://github.com/aspect-build/rules_lint/actions/runs/27764632302/job/82148156714?pr=914) looks like an infra / repo setup issue.

rafikk:
CI is green now. @jbedard, would you mind taking a look?

tools/sarif/sarif.go
```diff
@@ -179,6 +186,15 @@ func ToSarifJsonString(label string, mnemonic string, report string) (sarifJsonS
 		}
 	}
 
+	if mnemonic == "AspectRulesLintScalafix" {
```
jbedard, line 189: Why is this the first mnemonic we've had to do something like this for?

lint/scala/scalafix.bzl
```diff
@@ -0,0 +1,276 @@
...
+            If None, then fail the build when scalafix exits non-zero.
+        options: additional command-line arguments to scalafix
+        patch: output file for patch (optional). If provided, uses run_patcher for fix mode.
+        classpath: For semantic rules - transitive compile classpath (list of jars)
+    """
+    scala_toolchain = ctx.toolchains["@rules_scala//scala:toolchain_type"]
+
+    args = ctx.actions.args()
+    args.add_all(options)
+    args.add("--config", config.path)
```
jbedard: drop the `.path` and bazel will do that automatically?

tools/sarif/sarif.go
rafikk: I tried to explain this in the PR description, but let me give a detailed example. Linting [examples/scala/src/App.scala](https://github.com/rafikk/rules_lint/blob/cb281bc4dee7398469dc89c46ae48e303e54d9ec/examples/scala/src/App.scala), identifies 3 separate issues:

1. An unused import on line 1
2. Disabled procedure syntax on line 4
3. Disabled var syntax on line 5

The output from Scalafix for these findings looks like this:

```
/Users/rafiksalama/Library/Caches/bazel/_bazel_rafiksalama/00b67db09118a15005b2bd0b6fcee8c5/sandbox/darwin-sandbox/1/execroot/_main/src/App.scala:5:5: error: [DisableSyntax.var] mutable state should be avoided
    var foo = new Foo()
    ^^^
--- /Users/rafiksalama/Library/Caches/bazel/_bazel_rafiksalama/00b67db09118a15005b2bd0b6fcee8c5/sandbox/darwin-sandbox/1/execroot/_main/src/App.scala
+++ <expected fix>
@@ -1,7 +1,7 @@
-import java.net.URLDecoder

+
 object App {
-  def main(args: Array[String]) {
+  def main(args: Array[String]): Unit = {
     var foo = new Foo()
     println(foo.sayHello)
   }
```

The var syntax finding produces a great machine readable error: `src/App.scala:5:5: error: [DisableSyntax.var] mutable state should be avoided`. The `AspectRulesLintScalafix` errorfmt entries parses this output.

The other two issues however, are "rewritable", meaning that Scalafix has auto fix suggestions for them. When scalafix has autofixes, there is no way to get it to print the diagnostic. It only prints a patch with the changes. Hence, we get a diff that removes the unused import and replaces the procedure syntax.

`scalafixDiffEntries` is meant to process these diffs and produce a generic "Scalafix rewrite suggested" report on the first line of the diff.

All of this is not ideal. I've filed a Scalafix issue to discuss this at https://github.com/scalacenter/scalafix/issues/2459 and plan to upstream a command line flag that will print diagnostic lines for all errors when I have time.

If and when that's available, I think we should remove the diff entries parsing in `rules_lint`, require the minimum version of scalafix that includes the new flag, and pass the flag for the machine output rule. Until then, I think the current parser is a reasonable compromise.

lint/scala/lint/scalafix.bzl
```diff
@@ -0,0 +1,276 @@
+"""API for declaring a Scalafix lint aspect that visits scala_library, scala_binary, and scala test rules.
```
rafikk, line 1: @jbedard I followed the example for `aspect_rules_lint_rust` and placed `scalafix.bzl` at the root of the module so that it's imported using `load("@aspect_rules_lint_scala//:scalafix.bzl", "lint_scalafix_aspect")`

Given what we discussed, I wonder whether this should instead be `load("@aspect_rules_lint_scala//lint:scalafix.bzl", "lint_scalafix_aspect")` if we want to also move formatters into the language modules in the future. That way the language-specific modules also expose a `lint`/`format` hierarchy like the core ruleset.

tools/sarif/sarif.go
jbedard: Checkout `rubocopJsonToSarif()` and how that exception is done. Is that any better vs worse then how you've done it?

lint/scala/lint/scalafix.bzl
jbedard: I would vote for your suggestion... but it under a `lint` directory like the core.

lint/scala/MODULE.bazel
```diff
@@ -0,0 +1,15 @@
+"Bazel dependencies for Scala lint support"
+
+module(
+    name = "aspect_rules_lint_scala",
+    version = "0.0.0",
+)
+
+bazel_dep(name = "aspect_rules_lint", version = "2.7.1")
```
jbedard, line 8: FWIW due to the mess of `tools/sarif/sarif.go` containing everything in the world this will have to be incremented after the core changes are published before we first publish `aspect_rules_lint_scala`

lint/scala/lint/scalafix.bzl
rafikk: Done.

lint/scala/MODULE.bazel
rafikk: Yep. Understood. Once you've pushed a new `aspect_rules_lint` release, I'll bump the minimum version here and then we can publish the scala module.

tools/sarif/sarif.go
rafikk: It's similar in that both are processing custom formats and transforming them to SARIF. It is, however, worse, in the sense that it's not really necessary to do this in the Go tool. Since the output is JSON, it's trivial to preprocess or transform the rubocop output using JQ, similar to what I did https://github.com/aspect-build/rules_lint/pull/748

If the scalafix output was simpler to preprocess, I would have added an intermediate action before passing to the SARIF tool like the keep-sorted output. I may still do something similar in the future, but I would really prefer to just upstream the required fixes to scalafix, so that's why I didn't spend any time on it here.

jbedard (approved):

PR #916, opened 2026-06-18
fix(ty): unconditionally include main workspace bin directory in import search paths

udaya2899:
This PR unconditionally adds `ctx.bin_dir.path` to `import_paths` for `ty` aspect. 

### Why?

 Currently, lint/ty.bzl only populates import search paths from dep[PyInfo].imports or via dep.label.workspace_root (for external workspaces).

For standard generated targets in the main workspace (like a standard `py_proto_library` or `py_grpc_library` generated from `.proto` sources):
1. They do not specify an explicit imports attribute in their `PyInfo` provider.
2. `workspace_root` evaluates to "" (empty string).

This doesn't let ty to resolve generated files' import paths that don't have an `imports` attr.

 ```
   error[unresolved-import]: Import 'foo.bar.proto_name_py2' could not be resolved
```

---

### Test plan

- New test cases added

### DIFF
--- a/examples/python/test/BUILD
+++ b/examples/python/test/BUILD
@@ -29,6 +29,12 @@ write_file(
     content = ["import os"],
 )
 
+write_file(
+    name = "py_code_generator_no_imports",
+    out = "main_generated_no_imports.py",
+    content = ["import os"],
+)
+
 py_library(
     name = "generated_py",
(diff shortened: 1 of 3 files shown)


udaya2899:
Friendly ping @jbedard

jbedard (approved):

PR #943, opened 2026-07-17
refactor: only call clang tidy once

markww:
While trying to improve performance of a `clang-tidy` job in my project using `rules_lint`, I noticed that `clang_tidy_action` was called twice, once to produce human-readable output and once to produce machine output. This looks like, in the worst case, it could end up with `clang-tidy` being called twice per target. Perhaps it is the case that the cache will almost always save us, but it's not 100% guaranteed and it makes the code a bit harder to follow.

This change instead just calls `copy_file` to copy the human output to the machine output, which it can then alter to its hearts content without risking a second call out to `clang-tidy`.

---

### Test plan

This is a semantic noop, so existing test cases should cover it.

### DIFF
--- a/lint/clang_tidy.bzl
+++ b/lint/clang_tidy.bzl
@@ -37,13 +37,13 @@ clang_tidy = lint_clang_tidy_aspect(
 ```
 """
 
+load("@bazel_skylib//rules:copy_file.bzl", "copy_file")
 load("@bazel_skylib//rules/directory:providers.bzl", "DirectoryInfo")
 load("@bazel_tools//tools/build_defs/cc:action_names.bzl", "ACTION_NAMES")
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
 load("@rules_cc//cc:defs.bzl", "CcInfo", "cc_common")
 load("//lint/private:lint_aspect.bzl", "LintOptionsInfo", "OPTIONAL_SARIF_PARSER_TOOLCHAIN", "OUTFILE_FORMAT", "noop_lint_action", "output_files", "parse_to_sarif_action", "patch_and_output_files", "should_visit")
 load("//lint/private:patcher_action.bzl", "patcher_attrs", "run_patcher")
-load("@bazel_skylib//rules:copy_file.bzl", "copy_file")
 
 _MNEMONIC = "AspectRulesLintClangTidy"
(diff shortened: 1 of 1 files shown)


lint/clang_tidy.bzl
```diff
@@ -411,7 +412,8 @@ def _clang_tidy_aspect_impl(target, ctx):
 
         # TODO(alex): if we run with --fix, this will report the issues that were fixed. Does a machine reader want to know about them?
         raw_machine_report = ctx.actions.declare_file(OUTFILE_FORMAT.format(label = target.label.name + "_rules_lint/" + file.short_path, mnemonic = _MNEMONIC, suffix = "raw_machine_report"))
-        clang_tidy_action(ctx, compilation_context, ctx.executable, [file], raw_machine_report, output.machine.exit_code)
```
jbedard, line 414: Doesn't this output a completely different file format?

markww: I may be misunderstanding the code, but I don't see how it could be. `clang_tidy_action` is being called with the same arguments here as it is on line 402 other than output files and the possibility of patching. The only way the output format could be different is if some part of the infra someplace else is switching on the name of the output file(s) to decide the format, which would be pretty sketchy action-at-a-distance.

jbedard: Because the `clang_tidy_action()` seems identical for the machine vs human outputs, except the option `patch` argument. I think that means what you are suggesting is safe

Unrelated to your diff, but does that mean `parse_to_sarif_action()` is parsing human-readable output instead of proper sarif data for cland_tidy? 🤔

jbedard: But maybe you can just add a 1 line comment saying `clang_tidy_action()` is identical for human vs machine reports so we fallback to a simple `copy_file()`?

markww: > Unrelated to your diff, but does that mean `parse_to_sarif_action()` is parsing human-readable output instead of proper sarif data for cland_tidy? 🤔

I think that, for this case, clang tidy is producing *only* human readable output and `parse_to_sarif_action` is scraping that human readable output to produce sarif. I think there is no separate machine output that makes sense to produce for `clang-tidy` at this time.

> But maybe you can just add a 1 line comment saying clang_tidy_action() is identical for human vs machine reports so we fallback to a simple copy_file()?

Can do!

PR #949, opened 2026-07-25
feat(lint): support extra arguments for all linters

TrapsterDK:
AI-written, human-reviewed.

Add an `extra_args` parameter to every linter aspect factory and forward it to human, machine, fix, patch, and auxiliary actions where applicable. Existing Bandit and Clippy argument parameters remain supported.

Fixes #823.

---

All linter aspect factories now support additional command-line arguments through `extra_args`.

### Test plan

- Manual testing: analyzed the C++, Java, Kotlin, and QML examples with their lint configurations


lint/buf.bzl
```diff
@@ -21,7 +21,7 @@ _MNEMONIC = "AspectRulesLintBuf"
 def _short_path(file, _):
     return file.path
 
-def buf_lint_action(ctx, buf, protoc, target, stderr, exit_code = None):
+def buf_lint_action(ctx, buf, protoc, target, stderr, exit_code = None, options = []):
```
jbedard: Why is it called `args` / `extra_args` everywhere else but it's `options` in the actions? Shouldn't we just keep the "args" terminology?

TrapsterDK: It was basically done due to existing convention, quote from AI:
```
`options` to follow the existing convention for lower-level action helpers. In particular, `yamllint` and `scalafix` already expose `extra_args` in the aspect API and forward them to an action parameter named `options`:

- [`yamllint`](https://github.com/aspect-build/rules_lint/blob/7eced439085fa5acbc9ea3aed71bed587ec915b9/lint/yamllint.bzl#L102-L157)
- [`scalafix`](https://github.com/aspect-build/rules_lint/blob/7eced439085fa5acbc9ea3aed71bed587ec915b9/lint/scala/lint/scalafix.bzl#L200-L241)

Several other existing action helpers also use `options`, including `bandit_action`, `checkstyle_action`, `flake8_action`, `ktlint_action`, and `pylint_action`.
```

So I did it to avoid changing existing code, but if you want I can change it.

jbedard: If we're blindly passing through args I'd prefer keeping it as `args` or `extra_args`, even if it means changing 2 existing ones 🤷

jbedard (approved):

PR #950, opened 2026-07-25
feat: add SQLFluff linter

TrapsterDK:
AI-written, human-reviewed.

Add a SQLFluff lint aspect with human and deterministic SARIF output, standard patch generation, and automatic fixes.

The aspect supports SQLFluff 4.3.0 or newer and lets SQLFluff apply its native per-file configuration hierarchy when one Bazel target contains sources governed by different `.sqlfluff` files.

Templater dependencies can be declared globally on the aspect or locally through a visited target's `data` attribute. Target-local `data` is preferred for project-specific macros so unrelated lint actions do not receive those inputs.

SQL formatting remains unchanged in this PR. The formatter replacement is handled separately in #951.

---

### Changes are visible to end-users: yes

- Searched for relevant documentation and updated as needed: yes
- Breaking change (forces users to change their own code or config): no
- Suggested release notes appear below: yes

SQLFluff 4.3.0 or newer is now supported for SQL linting and automatic fixes, including nested configurations and target-local templater dependencies.

### Test plan

- Covered by existing test cases
- Added regression coverage for mixed nested configurations in one SQLFluff action
- Added target-local Jinja macro coverage and verified macros do not leak to sibling targets
- Validated human and deterministic SARIF output, hard failures, and fix patches
- Ran `bazel test //:requirements.test //test/...` in `examples/sql`
- Ran `bats test/lint_test.bats` in `examples/sql`
- Ran `bazel test //...` at the repository root

PR #951, opened 2026-07-25
feat(format): replace Prettier with SQLFluff for SQL

### LINKED ISSUE #947: Should SQLFluff be supported as an SQL formatter?
SQL formatting is currently provided by `prettier-plugin-sql`.

While working on SQLFluff linting support, I noticed that SQLFluff also provides a dedicated `sqlfluff format` command and supports a broad range of SQL dialects.

Before implementing formatter support, I wanted to ask which direction maintainers would prefer:

- Add SQLFluff as an alternative formatter while keeping `prettier-plugin-sql`.
- Replace or deprecate `prettier-plugin-sql` in favor of SQLFluff.
- Keep the existing formatter unchanged and add SQLFluff only as a linter.

I am happy to implement whichever approach best fits the project formatter design.

TrapsterDK:
AI-written, human-reviewed.

Replace `prettier-plugin-sql` with SQLFluff 4.3.0 or newer as the SQL formatter.

This PR is stacked on #950. Review the formatter commit `e568644`; the branch contains #950's squashed linter commit followed by one formatter commit.

This is a breaking change. Users must update the `sql` formatter binary from Prettier to SQLFluff. Generated targets are renamed from `format_SQL_with_prettier` to `format_SQL_with_sqlfluff`.

The SQL example now shares one SQLFluff binary and configuration between linting and formatting, and removes its Prettier SQL dependencies.

Closes #947.

---

SQL formatting now uses SQLFluff instead of `prettier-plugin-sql`.

### Test plan

- Added formatter integration coverage for fix and check modes
- Validated quiet output for successful and failing format checks
- Ran `bazel test //:requirements.test //test/...` in `examples/sql`
- Ran `bats test` in `examples/sql`
- Ran `bazel test //...` at the repository root
- Ran Buildifier checks on changed Starlark files


jbedard:
Isn't switching a breaking change? Should we not be simply adding opt-in support and we can switch the default in the next semver?

TrapsterDK:
Yes that is correct it is also written multiple times in the pr description, this is done based of the rule of there being one formatter per language. I can change the pr such that it just adds another formatter. (Never got a reply on the issue, mentioning possible implemention options and just chose to implement one as we just use rules_lint with a patch applied)

PR #956, opened 2026-08-03
fix: Patcher on windows should spawn .bat with cmd.exe

soda0289:
In nodejs if an executable ends in `.bat` or `.cmd` it can only be spawned as a `cmd.exe` argument or using `shell: true` spawn option. The linter executable can be a batch script `.bat` and needs to be launched with `cmd.exe`. 

This PR changes patcher.mjs to use `cmd.exe` to spawn batch scripts. Instead of explicitly calling `cmd.exe`, passing `/d /s /c`, and enabling `windowsVerbatimArguments` we could use nodejs `shell: true` option of spawnSync which does the same thing. The issue with `shell: true` is it leads to a deprecation warning [DEP0190](https://nodejs.org/api/deprecations.html#dep0190-passing-args-to-nodechild_process-execfilespawn-with-shell-option) if passed an array of arguments and might change behaviour with future versions of nodejs. I can change it to `shell: true` if want to keep the change smaller.

---

### Test plan

Linting should be working in the nodejs example with this patch applied on main branch. Must use bazel version >= 8.6 but not version 9. Bazel version 9 is not supported with the version of rules_swc in nodejs example.

```powershell
bazel --windows_enable_symlinks build --config=lint --enable_runfiles  --@aspect_rules_lint//lint:fix --output_groups=rules_lint_patch //src:ts_typings
```
Before:  throws an error
```
ERROR: .../src/BUILD:53:11: output 'src/ts_typings.AspectRulesLintESLint.patch' was not created
ERROR: Linting //src:ts_typings with ESLint failed: not all outputs were created or valid
```
After: it contains the fix for `const a: string = "a"`


jbedard:
This seems like multiple fixes that can be split into different PRs?

soda0289:
@jbedard Split the PR into 2 PRs. Second PR with fix for missing exe extension on diff binary is now #978

PR #959, opened 2026-08-04
fix: Set javascript module type to commonjs when using require()

soda0289:
When using rules_lint in an environment without a sandbox, such as windows, nodejs will determine the javascript module type (esm or cjs) using the nearest package.json which might exist in the users workspace/repo. When using a bazel sandbox package.json files will be hidden but if there is no sandbox other package.json files in the workspace/repo will be exposed and if they set type to esm/module then nodejs will fail to parse javascript files that use `require()`.

The eslint `.js` files are passed to eslint cli with `--format` option or to nodejs `--require` cli option. Since these are not run from js_binary directly it is not easy to copy a package.json file next to them. We could update eslint.bzl `_gather_inputs()` method to always include a package.json or could create another js_library for package.json and include it as a dep, but I felt this over complicated the fix. Changing the file extension to an  explicit `.cjs` is an easy fix. For rust library that uses js_binary we can just add package.json file to data attribute to ensure they use commonjs file type.

---

### Test plan

To test this create a workspace with a root package.json that sets `type: "module"` then run linting using eslint with no sandbox ` --spawn_strategy=local`. It will fail to run the eslint formatter since it uses `require()` when `.js` type is module.

Add  `"type": "module"`  to  `examples/nodejs/package.json`
```powershell
bazel build --config=lint --spawn_strategy=local --output_groups=rules_lint_human //src:ts_typings
```
Error:
```
ReferenceError: require is not defined in ES module scope, you can use import instead
This file is being treated as an ES module because it has a '.js' file extension and
'.../execroot/_main/package.json' contains "type": "module".
```


lint/rules_rust/BUILD.bazel
```diff
@@ -10,7 +10,10 @@ js_library(
 
 js_binary(
     name = "cli",
-    data = [":diagnostic-formatter"],
+    data = [
+        ":diagnostic-formatter",
+        "package.json",
+    ],
     entry_point = "rust.cli.js",
```
jbedard: Can this not be renamed instead of depending on `package.json`?

soda0289: We can rename them all. I was just worried about other developers adding new javascript files not using `.cjs` or .`mjs` file extension and breaking it again. I also feel like this package.json is not needed since it adds nothing of value, other than module type. I'll change the file extensions to be .cjs and try to add a failing test to prevent it from breaking in the future.

soda0289: I changed the files to cjs. I will stage up another PR that will test using eslint/nodejs examples without a sandbox. Using windows directly is probably better but need a few more patches merged to full test that.

jbedard (approved):

jbedard:
I would prefer we go with the explicit file extensions, then humans+robots will hopefully follow that convention going forward as well 👍

PR #971, opened 2026-08-14
fix(ty): preserve arguments in long materialized shell commands

### LINKED ISSUE #970: [Bug]: `lint_ty_aspect` fails when the generated shell command is too long
### What happened?

`lint_ty_aspect` fails when its generated shell command becomes long enough for Bazel to save it in a temporary script. The source and configuration arguments do not reach that script, so `ty` checks every Python file available to the action rather than the requested target.

### Version

Output of `bazel --version`: `7.6.0` or `9.2.0`.

`WORKSPACE` or `MODULE.bazel` file: latest `main` (4933d34a2c97518f6626b27657cf6841c3952222).

Language(s) and/or frameworks involved: Python with `ty`.

### How to reproduce

I have a reproduction committed on [`repro/ty-materialized-shell-args`](https://github.com/hartikainen/rules_lint/tree/repro/ty-materialized-shell-args) at `b14c8be8307d`. It adds a clean target with enough import roots to force command materialization and an unrelated transitive source with a known type error.

```console
$ cd examples/python
$ bazel test //test:ty_should_scope_materialized_command
//test:ty_should_scope_materialized_command FAILED
```

The test fails with the repository's Bazel `7.6.0` and with `USE_BAZEL_VERSION=9.2.0`.

#### Expected behavior

`uses_dependency_correctly.py` is clean, so its ty report should be empty and the test should pass.

#### Actual behavior

The report contains the diagnostic from the unrelated transitive source:

```console
error[unsupported-operator]: Unsupported `+` operation
 --> src/unsupported_operator.py:3:5
  |
3 | a = 10 + "test"
  |     --^^^------
```

`aquery` shows Bazel invoking the materialized script with the target arguments on the outer shell command:

```console
/bin/bash -c \
  bazel-out/.../uses_dependency_correctly_materialized.run_shell_0.sh \
  '' \
  src/uses_dependency_correctly.py \
  --project \
  .
```

The script itself contains:

```sh
ty check --force-exclude ... "$PARAM_FILE" $@
```

The source and `--project` arguments belong to the outer shell. The child script receives no arguments, so its `$@` is empty. The existing `ty_should_resolve_workspace_dep_with_imports` control test still passes because its shorter command remains inline.

### Any other information?

I discovered this in our repository with help from an LLM agent.

#### Possible fix

Avoid relying on `$@` across the materialized-script boundary. Embed the short source/config argument list into the generated script with shell quoting, or write it to a dedicated parameter file whose path is embedded in the script.

I tested the first approach locally. The reproduction test then passes, while `ty_should_fail_unsupported_operator` continues to report its expected diagnostic.

The same problem affects `aspect lint`, since Aspect CLI runs the same Bazel aspect actions and processes their output afterward.

hartikainen:
Closes #970.

For transparency: I both discovered and fixed this with help from an LLM agent.

`lint_ty_aspect` passes target sources and configuration through `$@`. That works while Bazel keeps the generated shell command inline.

Once enough import paths make the command long, Bazel materializes it as a child script. The arguments remain on the outer shell invocation, so the child script sees an empty `$@`. `ty` then checks every Python input available to the action rather than the requested target.

This PR:
- adds a regression that forces command materialization and demonstrates a clean target inheriting an unrelated transitive diagnostic.
- shell-quotes and embeds the short source/config argument list in the generated command.
- keeps the potentially long import search path in its existing parameter file.

The same fix applies to `aspect lint`, since Aspect CLI runs these Bazel aspect actions and processes their output afterward.

Fixed `lint_ty_aspect` losing target sources and configuration when Bazel materializes a long shell command. Large targets no longer lint unrelated transitive Python sources.

### Test plan

- New test cases added
- Manual testing:

  ```console
  cd examples/python

  # Repository default: Bazel 7.6.0
  bazel test //test:ty_should_scope_materialized_command

  # Bazel 9.2.0
  USE_BAZEL_VERSION=9.2.0 \
    bazel test //test:ty_should_scope_materialized_command
  ```

The new regression fails on the first commit and passes with the fix. All `10` ty-specific tests also pass.

### DIFF
--- a/lint/ty.bzl
+++ b/lint/ty.bzl
@@ -81,18 +81,24 @@ def ty_action(ctx, executable, srcs, transitive_srcs, config, stdout, exit_code 
         else:
             code_search_paths.append(p)
 
-    # Build a script that adds --extra-search-path only for directories that exist
-    # Some pip package directories may not exist, so we check first
-    # Pass --extra-search-path via a @param file, as there might be many of them
-    extra_search_path_script = """PARAM_FILE="$(mktemp)"
-"""
+    # Pass the candidate search paths in a Bazel-managed parameter file so the
+    # shell command stays constant-size. Some candidates may not exist, so the
+    # command filters them into the parameter file consumed by ty.
+    search_path_args = ctx.actions.args()
+    search_path_args.add_all(stub_search_paths + code_search_paths)
+    search_path_args.set_param_file_format("multiline")
+    search_path_args.use_param_file("%s", use_always = True)
 
-    for path in stub_search_paths + code_search_paths:
-        extra_search_path_script += """if [ -d "{path}" ]; then
-  echo "--extra-search-path" >> "$PARAM_FILE"
-  echo "{path}" >> "$PARAM_FILE"
-fi
-""".format(path = path)
+    extra_search_path_script = """readonly SEARCH_PATHS_FILE="$1"
+shift
+
+readonly PARAM_FILE="$(mktemp)"
+while IFS= read -r path; do
+  if [ -d "$path" ]; then
+    printf '%s\n' "--extra-search-path" "$path" >> "$PARAM_FILE"
+  fi
+done < "$SEARCH_PATHS_FILE"
+"""
 
     color_flag = "--color always" if color else "--color never"
     command = """
@@ -132,7 +138,7 @@ fi
             extra_search_path_script = extra_search_path_script,
             color_flag = color_flag,
         ),
-        arguments = [action_args],
+        arguments = [search_path_args, action_args],
         mnemonic = _MNEMONIC,
         env = env,
         progress_message = "Linting %{label} with ty",
--- a/examples/python/src/BUILD
+++ b/examples/python/src/BUILD
@@ -41,6 +41,26 @@ py_library(
     deps = ["//src/python_lib:py_dependency"],
 )
 
+# Forces Bazel to materialize the ty action's shell command. The transitive
+# source has a known type error but is not imported by the clean target below.
+py_library(
+    name = "many_imports_with_unrelated_error",
+    srcs = ["unsupported_operator.py"],
+    imports = [
+        "long_import_path_{}/subdirectory/that_makes_the_generated_shell_command_long".format(i)
+        for i in range(250)
+    ],
+)
+
+py_library(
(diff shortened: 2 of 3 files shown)


lint/ty.bzl
```diff
@@ -50,25 +53,25 @@ def ty_action(ctx, executable, srcs, transitive_srcs, config, stdout, exit_code
 
     # Wire command-line options, see
     # `ty help check` to see available options
-    action_args = ctx.actions.args()
```
jbedard, line 53: We need to keep using `ctx.actions.args`, especially while moving toward making rules_lint more bazel path-mapping friendly. That means using `ctx.actions.args` and never invoking `.path` (which already exists here, but you're making it worse.

I assume we need a different solution here...

hartikainen: Thanks for the quick response. I poked around a bit (with an agent, of course) and found a solution that looks reasonable to me.

The change is relatively simple and should be easy to review: https://github.com/aspect-build/rules_lint/compare/main...hartikainen:try/ty-bounded-shell-args. Basically, it just moves the list of import paths into a parameter file and replaces the generated per-path shell blocks with a constant-size loop.

Happy to amend this PR or open another one if that approach sounds reasonable.

hartikainen: I added the changes as a fixup to this branch. Let me know what you think.

lint/ty.bzl
```diff
@@ -132,7 +138,7 @@ fi
             extra_search_path_script = extra_search_path_script,
             color_flag = color_flag,
         ),
-        arguments = [action_args],
+        arguments = [search_path_args, action_args],
```
jbedard, line 141: So 1/2 of the `*_args` objects uses a file, and we pass a reference to that file via `SEARCH_PATHS_FILE`? I guess I almost always see a single `arguments` object when using `use_param_file` 🤷 

But if the tests pass (and your new tests failed before your fix) then LGTM 👍

hartikainen: Yeah, that's how I understand it.

Thanks for the review! I'll go ahead and squash the commits. Do I need to do something else to get this merged in?

jbedard: You can squash if you'd like, otherwise I'll squash on merge. I think nothing else is needed 👍

jbedard (approved):

PR #974, opened 2026-08-17
feat(ty): support passing type stubs via `types` attribute on `lint_ty_aspect`

faximan:
To make global type stubs (e.g. `types-*`, `*-stubs`) available to `ty`, they currently must be added to target `deps`. This pulls all stub packages and their transitive dependencies into runtime runfiles, which has the problematic side effect of bloating tarballs or containers that package such python targets, often with several 100 MB.

This PR fixes this problem by adding an optional `types` parameter to `lint_ty_aspect` that is only made available to ty and separated from `deps`. I also deduplicate `PyInfo` collection logic across `deps`, `srcs`, and `_types`.

#### Testing

In our downstream project, I
  * Verified `ty` resolves types and flags type errors using stubs provided via `types`.
  * Verified with `bazel cquery` and `du` that stub packages are omitted from the target's `.runfiles`.

### DIFF
--- a/lint/ty.bzl
+++ b/lint/ty.bzl
@@ -163,6 +163,20 @@ def _resolve_import_path(import_path, workspace_name, bin_dir_path):
     external_rel = "external/" + import_path
     return [external_rel, bin_dir_path + "/" + external_rel]
 
+def _collect_py_info(dep, ctx, transitive_sources, import_paths):
+    if PyInfo in dep:
+        transitive_sources.append(dep[PyInfo].transitive_sources)
+        transitive_sources.append(dep[PyInfo].transitive_pyi_files)
+        imports = dep[PyInfo].imports.to_list()
+        if imports:
+            for import_path in imports:
+                resolved = _resolve_import_path(import_path, ctx.workspace_name, ctx.bin_dir.path)
+                for e in resolved:
+                    import_paths[e] = True
+        elif dep.label.workspace_root:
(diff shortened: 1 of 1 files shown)


faximan:
@JesseTatasciore @jbedard I am getting "internal server error" trying to sign the CLA. Maybe this is related to GitHub's outage today, but right now they claim things are resolved and it still doesn't work for me.

Edit: resolved. It worked the next day.

faximan:
Friendly ping @JesseTatasciore @jbedard. Thanks for taking a look.

udaya2899 (approved):

faximan:
Friendly ping @JesseTatasciore @jbedard. Thanks for taking a look.

jbedard:
Can we get some tests or an example that actual use the new API?

faximan:
> Can we get some tests or an example that actual use the new API?

Thanks for taking a look! I added a test and example code.

jbedard:
The tests kind of point out a lack of functionality in rules_py[thon] more then anything. Shouldn't this be handled by the rulesets?

faximan:
Hmm I don't think so? How could rules_python differentiate between runtime deps and "meta deps" like these type stubs? If I am adding something to `deps` (as I am required to with rules_lint today), I would expect them to be available at runtime, and thus packaged into runfiles, given Python is not a compiled language.

PR #978, opened 2026-08-19
fix(patcher): Preserve the diff binary filename and extension

soda0289:
Windows will only execute files that have certain extensions such as `.exe` or `.bat`. To make diff command usable in patcher.mjs script it must have an `.exe` file extension.

This PR creates a new rule `_diff_bin` that creates a symlink of the diff binary from toolchain `@diff.bzl//diff/toolchain:execution_type` to bazel-out runfiles directory. I tried changing the current `genrule()` implementation but was not able to use `select()` on outs attribute or use the make variable expansion in outs variable to create a diff.exe in bazel output.

---

### Test plan

Linting should be working in the nodejs example with this patch applied on main branch. Must use bazel version >= 8.6 but not version 9. Bazel version 9 is not supported with the version of rules_swc in nodejs example.

```powershell
bazel --windows_enable_symlinks build --config=lint --enable_runfiles --@aspect_rules_lint//lint:fix --output_groups=rules_lint_patch //src:ts_typings
```
Before:  throws an error
```
ERROR: .../src/BUILD:53:11: output 'src/ts_typings.AspectRulesLintESLint.patch' was not created
ERROR: Linting //src:ts_typings with ESLint failed: not all outputs were created or valid
```
After: it contains the fix for `const a: string = "a"`

### DIFF
--- a/lint/private/patcher_binary.bzl
+++ b/lint/private/patcher_binary.bzl
@@ -1,9 +1,24 @@
-"""Workaround https://github.com/bazelbuild/bazel/issues/14009 to support Bazel prior to 8.3 & 9
+"Patcher binary that includes diff tool"
 
-TODO(3.0): when we drop support for Bazel 7 and 8.2, we can simplify this.
-"""
 load("@aspect_rules_js//js:defs.bzl", "js_binary")
-load("@bazel_features//:features.bzl", "bazel_features")
+
+_DIFF_TOOLCHAIN = "@diff.bzl//diff/toolchain:execution_type"
+
+def _diff_bin_impl(ctx):
+    diff_bin = ctx.toolchains[_DIFF_TOOLCHAIN].diffutilsinfo.diff_bin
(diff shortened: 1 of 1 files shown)


lint/private/patcher_binary.bzl
```diff
@@ -1,35 +1,39 @@
-"""Workaround https://github.com/bazelbuild/bazel/issues/14009 to support Bazel prior to 8.3 & 9
```
jbedard, line 1: Is this not still true?

soda0289: That issue and comment are about using toolchain_types with `genrule()`. The code has an if statment checking if that feature is supported `if bazel_features.toolchains.genrule_accepts_toolchain_types:` since it was only added in 8.3. If toolchain_types are not supported in `genrule()` then the rules_lint would not use diff from toolchain and fallback to diff from PATH. toolchain_types are supported with a normal `rule()` which this PR uses

jbedard (approved):

PR #979, opened 2026-08-20
fix(sarif): Use forward slash for artifactLocation.uri

soda0289:
SARIF's artifactLocation.uri is a URI and must always use a slash (/) separators. On windows file paths use backslash (\\) and will generate invalid uris. 

This PR changes the windows file path backslash to forward slash. It uses the `ToSlash()` method to convert backslash to forward slash, which only works on windows and is noop on linux. The tests cases use `FromSlash()` to convert unix path to windows path.

Stylelint already converts the [reported file path to unix paths](https://github.com/aspect-build/rules_lint/blob/cd20671330c5f85c703df5d4499011df37d55c3a/lint/js/stylelint.compactFormatter.mjs#L130) but other linters do not.

---

### Test plan

To test this use eslint on windows to generate sarif report. It will use invalid URI in artifactLocation.

```
    bazel test //test:eslint_machine_output_test.uri
```

Before: `"uri": "src\\file.ts"`
After:  `"uri": "src/file.ts`

### DIFF
--- a/tools/sarif/sarif_test.go
+++ b/tools/sarif/sarif_test.go
@@ -135,6 +135,19 @@ func TestSarif(t *testing.T) {
 		g.Expect(sarifJson.Runs[0].Results[0].Locations[0].PhysicalLocation.Region.GetRdfRange().Start.Line).To(Equal(int32(2)))
 	})
 
+	t.Run("determineRelativePath: converts Windows path separators to URI separators", func(t *testing.T) {
+		g := NewGomegaWithT(t)
+
+		// These run on every platform, so they exercise the conversion on Linux CI too.
+		g.Expect(determineRelativePath(`src\file.ts`, "//src:ts")).To(Equal("src/file.ts"))
+		g.Expect(determineRelativePath(`foo\bar\baz`, "//foo/bar:baz")).To(Equal("foo/bar/baz"))
+		g.Expect(determineRelativePath(`src\file.ts`, "")).To(Equal("src/file.ts"))
+
+		// A POSIX path is already a valid URI path and is returned untouched.
+		g.Expect(determineRelativePath("src/file.ts", "//src:ts")).To(Equal("src/file.ts"))
+		g.Expect(determineRelativePath("foo/bar/baz", "//foo/bar:baz")).To(Equal("foo/bar/baz"))
+	})
+
 	t.Run("determineRelativePath: returns relative paths untouched", func(t *testing.T) {
 		g := NewGomegaWithT(t)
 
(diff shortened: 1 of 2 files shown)


tools/sarif/sarif_test.go
```diff
@@ -157,6 +158,15 @@ func TestSarif(t *testing.T) {
 		g.Expect(determineRelativePath("foo/bar/baz", "//foo/bar:baz")).To(Equal("foo/bar/baz"))
 	})
 
+	t.Run("determineRelativePath: returns host path separators as URI separators", func(t *testing.T) {
+		g := NewGomegaWithT(t)
+
+        // Windows path with backslash. FromSlash and ToSlash only converts paths on windows
+		g.Expect(determineRelativePath(filepath.FromSlash("src/file.ts"), "//src:ts")).To(Equal("src/file.ts"))
```
jbedard: Shouldn't we the first argument as raw strings to ensure we are unit testing all types of paths, no matter which platform we are running on? When CI only mac/linux that means we are not testing your real fix atm?

I'd think we want...
```
		g.Expect(determineRelativePath("src/file.ts", "//src:ts")).To(Equal("src/file.ts"))
		g.Expect(determineRelativePath("src\file.ts", "//src:ts")).To(Equal("src/file.ts"))
```
?

soda0289: Yea I did write that originally the issue is that in go lang `ToSlash()` basically does `strings.replaceAll(input, os.PathSeperator, "/")` and `os.PathSeperator` is `/` on linux which means it won't actually convert the `\` to a `/` on linux. I could change `ToSlash()` to just a simple replace `\` with `/` but `\` is a valid path character on linux and would produce invalid paths if someone was using it in a filename. `ToSlash()` is doing the right thing where it only replaces `\` on windows and not linux.

If you like I can change it to just a basic find and replace of backslash to slash and not worry about the possibility of linux path unless there is some better way to handle it.

jbedard: Well if we're doing `filepath.FromSlash()` here then we're not really testing the fix, are we? Would this test fail before your change if run on windows?

soda0289: The tests didn't read very well. The issue is that the test need to be able to do windows file path manipulation on linux which the go standard library doesn't support. Using `FromSlash()` on linux won't do anything but on windows it will but we never test on windows.

I have modified the tests to use actual windows file paths. Both backslash file paths and windows absolute file paths with `c:\` are tested. I had to write methods to convert file paths and determine if they are absolute instead of using the built in go ones since they won't work under linux.

tools/sarif/sarif_test.go
```diff
@@ -181,6 +193,10 @@ func TestSarif(t *testing.T) {
 		g.Expect(determineRelativePath("/mnt/ephemeral/output/bazel-examples/__main__/sandbox/linux-sandbox/769/execroot/_main/speller/lookup/lookup-test.cc", "//:lookup")).To(Equal("speller/lookup/lookup-test.cc"))
 		g.Expect(determineRelativePath("/mnt/ephemeral/output/bazel-examples/__main__/sandbox/linux-sandbox/780/execroot/_main/speller/data_driven_tests/lookup-datatest.cc", "//:data_driven_tests")).To(Equal("speller/data_driven_tests/lookup-datatest.cc"))
 		g.Expect(determineRelativePath("/private/var/tmp/_bazel_jesse/93d7e699c5e2019d94351d19b00be5a3/sandbox/darwin-sandbox/249/execroot/_main/speller/announce/announce.cc", "//:announce")).To(Equal("speller/announce/announce.cc"))
+
+		// Windows path, with a drive letter instead of a leading slash
+		g.Expect(determineRelativePath("c:/private/var/tmp/_bazel_jesse/93d7e699c5e2019d94351d19b00be5a3/sandbox/darwin-sandbox/249/execroot/_main/speller/announce/announce.cc", "//speller/announce:announce")).To(Equal("speller/announce/announce.cc"))
+		g.Expect(determineRelativePath("C:/users/jesse/_bazel/abc123/execroot/_main/speller/announce/announce.cc", "//:announce")).To(Equal("speller/announce/announce.cc"))
```
jbedard, line 201: Shouldn't we also have tests with `\`?

soda0289: I added the test cases with backslash and forward slash. Had to just do a simple find and replace backslash to forward slash instead of using OS path separator since on linux test run it would not convert these paths.

jbedard:
@codex

jbedard (approved):

PR #287, opened 2024-06-15
Feat: clang-tidy for linting of C/C++ code

peakschris:
Support clang-tidy for linting of C++ code.

Inspired by and containing some code from bazel_clang_tidy repo. This integration has been discussed by maintainers of both repos in https://github.com/erenon/bazel_clang_tidy/issues/35.

This PR replaces #284, which had become noisy with too many commits. Since then, the code has been rewritten to return a single report for each target.

Fixes #178 

---

### Changes are visible to end-users: yes

- Searched for relevant documentation and updated as needed: yes
- Breaking change (forces users to change their own code or config): no
- Suggested release notes appear below: todo

### Test plan

- New test cases added
- Extra test cases could be added but need discussion; what is possible in CI?

### Code used from bazel_clang_tidy repo

bazel_clang_tidy repo helped:
- two helper functions toolchain_flags and safe_flags
- some lines in get_args
- inspiration & learning

The bulk of the code is implemented by following existing rules_lint code such as ruff.bzl and eslint.bzl.

### Input into release notes (will reformat)
- warning that git bash does not handle arguments correctly and --fix may have issues on Windows. msys64 bash (already recommended by Bazel is ok).

PR #985, opened 2026-08-23
fix(clang-tidy): stop shell-quoting -header-filter

### LINKED ISSUE #984: [Bug]: clang-tidy -header-filter arrives shell-quoted since v2.8.0, failing every lint_target_headers aspect
### What happened?

A `lint_clang_tidy_aspect` that sets `lint_target_headers = True` (or `header_filter = "..."`) fails to build its lint targets. Any target with headers is affected:

```
ERROR: .../src/cpp/lib/BUILD:7:11: Linting //src/cpp/lib:hello-time:get-time.cc
with clang-tidy failed: (Exit 1)
  .../clang_tidy '"-header-filter=.*src[\/]cpp[\/]lib/.*"' ...

error: no such file or directory: '.../"-header-filter=..."' [clang-diagnostic-error]
```

`-header-filter` arrives with literal `"` characters around the whole argument, so clang-tidy does not recognize it as a flag and treats it as an input path. That is a `clang-diagnostic-error`, which the wrapper treats as fatal. It affects both consumers: the `run_shell` lint path, and the `--fix` path, where `patcher.mjs` uses `spawnSync` without a shell and passes the quotes through unchanged.

v2.7.2 is fine and the likely cause of v2.8.0 regression is #867

### Version

Development (host) and target OS/architectures: linux x86_64

Output of `bazel --version`: bazel 9.2.0 (also tested on bazel 7.6.0)

Version of the Aspect rules, or other relevant rules from your `WORKSPACE` or `MODULE.bazel` file: aspect_rules_lint v2.8.0, and current main

Language(s) and/or frameworks involved: C++ / clang-tidy (LLVM 16.0.0 as pinned by examples/cpp; also reproduced against LLVM 21.1.8)

### How to reproduce

```shell
# IMPORTANT: needs a host where the pinned LLVM 16.0.0 can actually load,
# i.e. libtinfo.so.5 present. On ubuntu-24.04 clang-tidy fails to start and
# the wrapper reports that as success, so this command goes GREEN and the
# bug is masked -- that is the same reason CI does not catch it, below.
# Use ubuntu-22.04, or `apt-get install libtinfo5`, or first bump
# examples/cpp/MODULE.bazel to "16.0.4".
# Reproduced on a linux x86_64 host that has libtinfo.so.5 present.

# examples/cpp sets lint_target_headers = True
# (tools/lint/linters.bzl:13 and :32), so main reproduces with the
# exact command the `lint` job runs:
cd examples/cpp
bazel build --config=lint --output_groups=rules_lint_human //...

# Expected: lint targets build, reports contain clang-tidy findings.
# Actual:   Exit 1, clang-diagnostic-error, the action fails to build.
# Confirmed on a clean --output_base, so it is not a stale-cache artifact.
# Note the --output_groups=rules_lint_patch --@aspect_rules_lint//lint:fix
# variant does NOT fail: js_binary captures the exit code there.
```

#### Root cause

`_quoted_arg` in `lint/clang_tidy.bzl` wraps the value in quotes:

```starlark
def _quoted_arg(arg):
    return "\"" + arg + "\""
```

It dates from #287, which introduced clang-tidy support together with a wrapper that built one command string and `eval`'d it:

```bash
command="$clang_tidy $@ $file > $raw_out_file 2>&1"
eval $command
```

The shell removed those quotes on re-parse, so they were invisible. #867 correctly replaced the `eval` with a direct `"$clang_tidy" "$@"` invocation, so arguments now survive verbatim — and so do the quotes. No other linter is affected: #867 touched only `clang_tidy_wrapper.bash` and its test, and `_quoted_arg` had no other caller.

#### Why CI does not catch this

The `lint (examples/cpp)` job passes because its clang-tidy exits before it runs, and the wrapper cannot tell that apart from a clean lint. The example pins LLVM 16.0.0, whose linux-x86_64 tarball is built on ubuntu-18.04 and links `libtinfo.so.5`, which 24.04 does not ship. The `format` jobs carry a workaround — ubuntu-22.04 plus an apt install; the `lint` jobs, introduced on flat ubuntu-24.04 in #679, don't have it.

The loader failure then goes unnoticed: it carries no `clang-diagnostic-error`, so the wrapper records the exit code in `CLANG_TIDY__EXIT_CODE_OUTPUT_FILE`, returns success, and writes the message to the report file rather than stdout.

#### Workaround

Stay on v2.7.2 
or patch `_quoted_arg`'s body to `return arg` via `single_version_override` 
or drop `lint_target_headers` / `header_filter` and pass the flag through the `args` parameter added in #949

#### Suggested fixes

1. Delete `_quoted_arg` and append `-header-filter=` plainly.
2. Make the example's clang-tidy loadable, so CI covers this path. Bumping `examples/cpp` to LLVM 16.0.4 (built on 22.04, links `libtinfo6`) is a net deletion: the two `format` jobs return to plain `ubuntu-24.04` like the `lint` ones, and both apt steps go.
3. Stop the wrapper reporting success when clang-tidy could not run at all. Exit 126/127 mean the binary never executed, so there are no findings to capture. This is what let a broken toolchain read as a clean lint, independently of the bug above.

vadikmironov:
Fixes #984.

`_quoted_arg` wrapped `-header-filter` in literal `"` characters. That only worked while the wrapper `eval`'d its command string (#287). #867 replaced the `eval` with a direct `"$@"` invocation, so since v2.8.0 the quotes reach clang-tidy verbatim, it reads the flag as an input path, and every `lint_clang_tidy_aspect` with `lint_target_headers` or `header_filter` fails its lint actions with `clang-diagnostic-error`. v2.9.0 is affected too.

This PR only removes the quoting.

- Searched for relevant documentation and updated as needed: nothing documents the quoting

- Fixed `-header-filter` reaching clang-tidy wrapped in literal quotes, which made every `lint_clang_tidy_aspect` using `lint_target_headers` or `header_filter` fail to build its lint targets since v2.8.0.

### Test plan

- Manual, on a host where the pinned LLVM 16.0.0 can load (ubuntu-22.04, or `apt-get install libtinfo5`): in `examples/cpp`, `bazel build --config=lint --output_groups=rules_lint_human //...` goes from `clang-diagnostic-error` to clean, with real findings in the reports. Reverting this commit turns it red again.
- CI cannot exercise this path yet: on the ubuntu-24.04 lint runners the pinned clang-tidy exits 127 before running and the wrapper records that as a clean lint. Two follow-ups after this lands, in order: bump `examples/cpp` to LLVM 16.0.4 (`MODULE.bazel` only), then make the wrapper treat exit 126/127 as fatal. Each turns `lint (examples/cpp)` red without its predecessor.

This work is produced with the help of an LLM agent, reviewed by myself.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -50,8 +50,7 @@ jobs:
           echo "Generated test matrix: $folders"
   format:
     needs: list-examples
-    # Our version of clang-format has some deps that aren't satisfiable on newer ubuntu yet
-    runs-on: ${{ matrix.folder == 'examples/cpp' && 'ubuntu-22.04' || 'ubuntu-24.04' }}
+    runs-on: ubuntu-24.04
     strategy:
       fail-fast: false
       matrix:
@@ -64,9 +63,6 @@ jobs:
             common     --show_progress_rate_limit=60
             common     --show_timestamps
             build:lint --no@aspect_rules_lint//lint:color
-      # Permit this gross non-hermetic dependency
-      - if: matrix.folder == 'examples/cpp'
-        run: sudo apt-get update && sudo apt-get install -y libtinfo5
       # TODO: Fix for Bazel 9
       - if: matrix.folder == 'examples/csharp' || matrix.folder == 'examples/fsharp' || matrix.folder == 'examples/nodejs'
         run: echo "USE_BAZEL_VERSION=8.x" >> "$GITHUB_ENV"
@@ -80,8 +76,7 @@ jobs:
   # example sources are intentionally unformatted), matching the `format` job.
   format-aspect-cli:
     needs: list-examples
-    # Our version of clang-format has some deps that aren't satisfiable on newer ubuntu yet
-    runs-on: ${{ matrix.folder == 'examples/cpp' && 'ubuntu-22.04' || 'ubuntu-24.04' }}
+    runs-on: ubuntu-24.04
     strategy:
       fail-fast: false
       matrix:
@@ -91,9 +86,6 @@ jobs:
       - uses: aspect-build/setup-aspect@b1e9d142e86f63c10d603304daa22bf80492e24c # v2026.25.1
         with:
           aspect-api-token: ${{ secrets.ASPECT_API_TOKEN }}
-      # Permit this gross non-hermetic dependency
-      - if: matrix.folder == 'examples/cpp'
-        run: sudo apt-get update && sudo apt-get install -y libtinfo5
       # TODO: Fix for Bazel 9
       - if: matrix.folder == 'examples/csharp' || matrix.folder == 'examples/fsharp' || matrix.folder == 'examples/nodejs'
         run: echo "USE_BAZEL_VERSION=8.x" >> "$GITHUB_ENV"
--- a/examples/cpp/MODULE.bazel
+++ b/examples/cpp/MODULE.bazel
@@ -16,7 +16,9 @@ local_path_override(
 llvm = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm")
 llvm.toolchain(
     llvm_versions = {
-        "": "16.0.0",
+        # Keep to a release whose linux tarball is built on ubuntu-22.04 —
+        # 16.0.0 and 18.1.4 are built on 18.04 and link libtinfo5
+        "": "16.0.4",
         "darwin-x86_64": "15.0.7",
     },
 )
--- a/lint/clang_tidy.bzl
+++ b/lint/clang_tidy.bzl
@@ -231,9 +231,6 @@ def _aggregate_regex(ctx, compilation_context):
         print("target header dirs: " + ",".join(dirs))
     return regex
 
-def _quoted_arg(arg):
-    return "\"" + arg + "\""
-
 def _get_env(ctx, srcs):
     sources_are_cxx = _is_cxx(srcs[0])
     if (sources_are_cxx):
@@ -251,16 +248,18 @@ def _get_env(ctx, srcs):
     return env
 
 def _get_args(ctx, compilation_context, srcs):
+    # These reach clang-tidy verbatim; the wrapper runs it directly, so shell
+    # quoting added here arrives as literal characters.
     args = []
     if (any(ctx.files._global_config)):
         args.append("--config-file=" + ctx.files._global_config[0].path)
     if (ctx.attr._lint_target_headers):
         regex = _aggregate_regex(ctx, compilation_context)
         if (regex):
-            args.append(_quoted_arg("-header-filter=" + regex))
+            args.append("-header-filter=" + regex)
     elif (ctx.attr._header_filter):
         regex = ctx.attr._header_filter
-        args.append(_quoted_arg("-header-filter=" + regex))
+        args.append("-header-filter=" + regex)
     args.extend([src.path for src in srcs])
     return args
 
--- a/lint/clang_tidy_wrapper.bash
+++ b/lint/clang_tidy_wrapper.bash
@@ -63,7 +63,12 @@ if [[ -z $CLANG_TIDY__STDOUT_STDERR_OUTPUT_FILE ]]; then
 fi
 # distinguish between compile (fatal) errors and warnings-as-errors errors
 fatal_error=0
-if [ $exit_code -ne 0 ] && [ -s $out_file ]; then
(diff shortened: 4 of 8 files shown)


.github/workflows/ci.yaml
```diff
@@ -50,8 +50,7 @@ jobs:
           echo "Generated test matrix: $folders"
   format:
     needs: list-examples
-    # Our version of clang-format has some deps that aren't satisfiable on newer ubuntu yet
```
jbedard, line 53: Why do this as part of this PR?

vadikmironov: This part is not a required part of the change strictly speaking and can run as a separate PR, but the choices to make a fix to clang_tidy_wrapper with a working CI were:
- either replicate the non-hermetic libtinfo5 dependency for the lint job
- or bump LLVM to a version without this dependency (ok, to be precise swapping libtinfo5 depedency in favour of libtinfo6 which happens to be present in ubuntu-24.04) 

However, part of the reason CI for lint was broken is the fact that formatting and linting, despite using exactly the same toolchain, are configured with different runners. The line `Permit this gross non-hermetic dependency` sort of tipped my thinking towards not copy-pasting it for the lint job, and instead clean this particular line and the formatting job to be exactly the same as lint instead. Happy to do it the other way and copy below block to the lint job if that's preferred:

```
      # Permit this gross non-hermetic dependency
      - if: matrix.folder == 'examples/cpp'
        run: sudo apt-get update && sudo apt-get install -y libtinfo5
```

jbedard, line 53: I think nothing from this file should be changing?

vadikmironov: Apologies, on Sept 3 I hit "Start a review" instead of "Add single comment", so my reply sat as a pending draft. Here it is in full:

> This part is not a required part of the change strictly speaking and can run as a separate PR, but the choices
> to make a fix to clang_tidy_wrapper with a working CI were:
> - either replicate the non-hermetic libtinfo5 dependency for the lint job
> - or bump LLVM to a version without this dependency (ok, to be precise swapping libtinfo5 dependency in 
> favour of libtinfo6 which happens to be present in ubuntu-24.04)
>
> However, part of the reason CI for lint was broken is the fact that formatting and linting, despite using 
> exactly the same toolchain, are configured with different runners. The line `Permit this gross non-hermetic 
> dependency` sort of tipped my thinking towards not copy-pasting it for the lint job, and instead clean this 
> particular line and the formatting job to be exactly the same as lint instead. Happy to do it the other way 
> and copy below block to the lint job if that's preferred:
>
> ```
>       # Permit this gross non-hermetic dependency
>       - if: matrix.folder == 'examples/cpp'
>         run: sudo apt-get update && sudo apt-get install -y libtinfo5
> ```

Given that you haven't seen my original response (my bad, should've checked), I would like to confirm if you want ci.yaml to be left alone entirely including format job. If so I will revert the ci.yaml part, but will need to keep the 16.0.4 bump in examples/cpp/MODULE.bazel. The lint job already runs on ubuntu-24.04 and clang-tidy is going to be fine there. The format jobs is working on ubuntu-22.04 and I think jammy ships libtinfo6. 

Let me know if this is an acceptable path forward.

jbedard: I'd prefer if the entire diff was only "stop shell-quoting -header-filter" related, if there are issues with formatting or commit hooks doing extra things please don't do them or do them in a separate PR.

vadikmironov: I reduced this PR to `_quoted_arg` removal only, rebased on main.

To be clear though: nothing in the original diff was formatter or hook doing extra things. Each part is there for a reason, explained in #984 and in this PR's description. A quick recap of the three stacked changes:

1. The actual fix for `_quoted_arg`, which is all this PR is now.
2. Make CI actually run clang-tidy. `lint (examples/cpp)` has been permagreen since #679: the pinned LLVM 16.0.0 links `libtinfo5`, absent on ubuntu-24.04, so clang-tidy aborts and the wrapper records that as a clean lint. That is how #867 broke every `header_filter` / `lint_target_headers` user with green CI and it shipped in v2.8.0 and again in v2.9.0. Locally, wherever clang-tidy can load, `examples/cpp` fails on main today.
3. Make the wrapper fail on "never ran" (exit 126/127), so the next clang-tidy with missing dependencies in CI fails instead of reporting a clean lint.

They are of course separable, and they are now separate. I can stop after 1 if you still think 2 and 3 not required. This would leave CI broken and blind to this class of regression, but if this is a preference it's fine by me.

jbedard (approved):

PR #986, opened 2026-08-26
feat(cppcheck): build SARIF output

martin4861:
The cppcheck aspect ran `cppcheck` three times: human, a separate text run parsed into SARIF, and xml. The text run did not produce usable SARIF.

This change runs `cppcheck` twice (human and xml):
- The human run pins `--template={file}:{line}:{column}: {severity}: {message} [{severity} {id}]` so the layout the SARIF parser expects is guaranteed. Its output is directly used as input for the SARIF parser, which has been updated to this format.
- xml output stays and is unchanged

Additional fixes to make the export work:
- `_prefixed` emitted `-I dir` instead of `-Idir`, so include paths were passed as two `argv` entries
- The xml run writes via `--output-file` instead of a stdout redirect, and the file is created up-front because cppcheck writes nothing when it fails before checking starts.
- The aspect now honors `should_visit`, with a new `rule_kinds` parameter on `lint_cppcheck_aspect` defaulting to `["cc_binary", "cc_library"]`.

The `cppcheck` aspect now emits valid SARIF and runs one fewer action per target. 
`lint_cppcheck_aspect` accepts `rule_kinds` (default `["cc_binary", "cc_library"]`) to limit which target kinds are visited.

### Test plan

- Test cases for the SARIF exporter
- Manual tests in `examples/cpp` and our internal code base

### DIFF
--- a/lint/cppcheck.bzl
+++ b/lint/cppcheck.bzl
@@ -37,7 +37,7 @@ def _filter_srcs(rule):
 def _prefixed(list, prefix):
     array = []
     for arg in list:
-        array.append("{} {}".format(prefix, arg))
+        array.append("{}{}".format(prefix, arg))
     return array
 
 def _get_compiler_args(compilation_context):
--- a/lint/cppcheck.bzl
+++ b/lint/cppcheck.bzl
@@ -3,7 +3,7 @@
 
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
 load("@rules_cc//cc:defs.bzl", "CcInfo", "cc_common")
-load("//lint/private:lint_aspect.bzl", "LintOptionsInfo", "OPTIONAL_SARIF_PARSER_TOOLCHAIN", "OUTFILE_FORMAT", "noop_lint_action", "output_files", "parse_to_sarif_action")
+load("//lint/private:lint_aspect.bzl", "LintOptionsInfo", "OPTIONAL_SARIF_PARSER_TOOLCHAIN", "OUTFILE_FORMAT", "noop_lint_action", "output_files", "parse_to_sarif_action", "should_visit")
 
 _MNEMONIC = "AspectRulesLintCppCheck"
 
@@ -101,6 +101,9 @@ def cppcheck_action(ctx, compilation_context, executable, srcs, stdout, exit_cod
     )
 
 def _cppcheck_aspect_impl(target, ctx):
+    if not should_visit(ctx.rule, ctx.attr._rule_kinds):
+        return []
+
     if not CcInfo in target:
         return []
 
@@ -138,7 +141,9 @@ def _cppcheck_aspect_impl(target, ctx):
     )
     return [info]
 
-def lint_cppcheck_aspect(binary, verbose = False, args = []):
+DEFAULT_RULE_KINDS = ["cc_binary", "cc_library"]
+
+def lint_cppcheck_aspect(binary, verbose = False, rule_kinds = DEFAULT_RULE_KINDS, args = []):
     """A factory function to create a linter aspect.
 
     Args:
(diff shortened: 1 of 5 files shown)


lint/cppcheck.bzl
```diff
@@ -101,6 +101,9 @@ def cppcheck_action(ctx, compilation_context, executable, srcs, stdout, exit_cod
     )
 
 def _cppcheck_aspect_impl(target, ctx):
+    if not should_visit(ctx.rule, ctx.attr._rule_kinds):
```
jbedard: Can we do the `CcInfo` check first? Just because I assume it's significantly cheaper...

tools/sarif/cppcheck.go
```diff
@@ -0,0 +1,206 @@
...
+	"encoding/json"
+	"encoding/xml"
+	"fmt"
+	"strings"
+)
+
+const sarifSchema = "https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/schemas/sarif-schema-2.1.0.json"
+const sarifVersion = "2.1.0"
+
+// cppcheck `--xml-version=3` document, see https://cppcheck.sourceforge.io/manual.pdf
```
jbedard: Any idea if `cppcheck` if following any standard XML output format here or is it something they invented? Maybe if they're following a standard we can use this code for something other then `cppcheck` in the future...?

martin4861: I was able to fix the xml output so that it follows a standard xml output format. I still have to keep `cppcheck.go` (but could simplify it), so that we are able to convert `cppcheck`'s xml output to SARIF (for GHA annotations).

(In theory, we could also create the SARIF report out of the human-readable output. But doing this would require us to run `cppcheck` twice, because we always need the xml output to generate the MISRA compliant output. This is a post processing step using the premium tools of `cppcheck`)

jbedard: So this XML is "MISRA compliant"? Can this golang code be used for other tools that output "MISRA compliant output"?

lint/extensions.bzl
```diff
@@ -20,21 +20,21 @@ def _public_build_file_content(line):
...
         for r in mod.tags.cppcheck:
-            # Download cppcheck premium tar files for different platforms
-            # Even though cppcheckpremium is downloaded, it behaves like the free version, if we do not
-            # provide a cppcheck.cfg file and license key.
-
+            # cppcheck premium behaves like the free version and does not require a license,
+            # if cppcheck.cfg is removed.
             http_archive(
                 name = r.linux,
-                build_file_content = _public_build_file_content("""filegroup(name = "files", srcs = glob(["**"], exclude = ["cppcheck.cfg"]))"""),
```
jbedard: This seems unrelated to your main change? What is this? Can it be in a separate PR?

martin4861: That's a fix for a bug, which I found when running the linters in `examples/cpp`. I extracted it into https://github.com/aspect-build/rules_lint/pull/994 and will remove this commit from this PR.

tools/sarif/cppcheck.go
martin4861: No, it's not MISRA compliant. In order to keep the implementation easy and to avoid having the xml parser, I am now using the properly formatted human readable output, which can be easily parsed (similar to other linters). 

(Note: cppcheck allows to create a SARIF report, but it is missing important information, which we need for the GHA annotations)

jbedard: Ok, that removes the cpp specific golang file which is what I was worried about :+1:

jbedard (approved):

jbedard:
@martin4861 can you update the PR description? I think it's out of date now?

martin4861:
> @martin4861 can you update the PR description? I think it's out of date now?

Thanks for your review and your time. I updated the PR description accordingly!

PR #998, opened 2026-09-07
fix(rust): Remove toolchain selection from MODULE.bazel

gergondet-woven:
When consuming rules_rs downstream, one cannot re-use the recommended way to select the default toolchain version, i.e.:

```bzl
toolchains = use_extension("@rules_rs//rs/toolchains:module_extension.bzl", "toolchains")
toolchains.toolchain(
    edition = "2024",
    version = "1.92.0",
)
use_repo(toolchains, "default_rust_toolchains")
```

If this call doesn't specifically match the settings in aspect_rules_lint_rust, the downstream user will get an error message about conflicting versions for `default_rust_toolchains` which is not ideal.

By removing that call, rules_lint will:
- use whatever version rules_rs selects as default (typically in the examples/rust case)
- pick the version decided downstream

---

### Test plan

- Manual testing: use aspect_rules_lint_rust in a downstream project that attempt to setup a different default rust toolchain


gergondet-woven:
I think the CI failure is related to not having: https://github.com/aspect-build/rules_lint/pull/996

jbedard:
I've merged https://github.com/aspect-build/rules_lint/pull/996, can you rebase this one to make sure it's green since this is also rust related?

jbedard (approved):

jbedard:
Nevermind, I confirmed myself that fixed it, thanks! 👍

PR #999, opened 2026-09-07
fix(ruff): Make sure multiple mnemonic can co-exist

gergondet-woven:
This is following up on #952 which was missing one bit. I thought this was handled by passing the mnemonic to run_patcher but evidently not :pray: 

When using multiple ruff aspect with different mnemonic, you would get conflicts due to different aspect outputting the same patch file.

This PR fixes that and allows multiple mnemonics to co-exist peacefully.

The generated patch name is unchanged from the default when using the default mnemonic to keep it 100% transparent.

---

### Test plan


lint/ruff.bzl
```diff
@@ -98,6 +98,7 @@ def ruff_action(ctx, executable, srcs, config, stdout, exit_code = None, env = {
             args = action_args,
             files_to_diff = [s.path for s in srcs],
             patch_cfg_env = env,
+            patch_cfg_suffix = "{}patch_cfg".format("." + mnemonic if mnemonic != _MNEMONIC else ""),
```
jbedard: I think the way `patch_cfg_suffix` is used it will do the `"."` prefix for you and you can drop the `if/else`? Or you can pass `None` if `mnemonic = _MNEMONIC` to prevent changes in the default case?

gergondet-woven: Sorry, I had the mnemonic and dot order mixed up, I have switched them so it's now `{mnemonic}.patch_cfg` if the mnemonic changes from the default

> Or you can pass None if mnemonic = _MNEMONIC to prevent changes in the default case?

In the current implementation, `run_patcher` doesn't handle the case when `patch_cfg_suffix` is None but I could change this. Let me know what you prefer :saluting_face:

jbedard (approved):

PR #1002, opened 2026-09-15
feat(aspect): send builds to Aspect Cloud from CI

joshcbarnes:
Connects this repo's builds to Aspect Cloud's remote cache and BES, so invocations appear in `app.aspect.build`.

### The gap this closes

This repo's CI runs on **stock GitHub Actions runners, not Aspect Workflows runners**. That matters more than it looks: the Aspect CLI wires BES from the environment (`ASPECT_WORKFLOWS_BES_BACKEND` and friends, set by a Workflows runner). None of those variables exist here, and nothing passed `--bes_backend` or `--remote`, so **the CLI tasks were uploading nowhere**.

Easy to miss, because the Aspect GitHub App's check runs travel a *different* path — the API, not BES. Status checks looked healthy on every PR while zero invocations reached the UI.

### `.aspect/bazelrc` (new)

Follows the pattern from aspect-build/aspect-cli#1442. Flags are grouped by the endpoint that justifies them, so a deployment with a cache but no BES gets the cache tuning and none of the BES tuning:

| Group | Presumes | Holds |
|---|---|---|
| `aspect-common` | nothing | `--heap_dump_on_oom`, `--experimental_repository_cache_hardlinks` |
| `aspect-cache` | a remote cache | `--remote_upload_local_results`, `--remote_accept_cached`, `--remote_timeout`, `--remote_retries`, `--grpc_keepalive_timeout` |
| `aspect-bes` | a BES backend | profile generation + enrichment, `--legacy_important_outputs`, `--remote_build_event_upload=all` |
| `aspect-execlog` | a writable `.aspect/generated` | `--execution_log_compact_file` |

`--config=aspect-cloud` chains the first three and adds the endpoints plus the `aspect` credential helper, scoped per host. Every line is `common:<config>`-scoped, so the file is **inert until named** — a fresh clone builds exactly as before.

`aspect-execlog` is deliberately *not* in `aspect-cloud`. Its path is `%workspace%`-relative and Bazel will not create the parent directory, so folding it back in breaks all 26 example workspaces. Only the root workspace names it.

`.bazelrc` gains `try-import %workspace%/.aspect/bazelrc` at the top, so the repo's own lines override.

### One mechanism, every job

`.aspect/config.axl` can only ever cover `aspect-test`: it is read by the Aspect CLI alone *and* is per-workspace, so it reaches neither the vanilla-`bazel` jobs nor the CLI jobs that run in `examples/*`. The repo's `.bazelrc` has the same workspace limit. The one location indifferent to both axes is the **user** bazelrc, `$HOME/.bazelrc`, which every `bazel` process on the runner reads regardless of directory — and which the Aspect CLI also re-injects into the Bazel it spawns (verified on the `test`, `format` and `lint` tasks).

So a single composite action, `.github/actions/aspect-cloud/`, writes that file and every job family uses it. `config.axl` keeps only `--config=aspect-execlog`, which is genuinely workspace-bound: its path is `%workspace%`-relative and only the root workspace has `.aspect/generated`.

This matters beyond tidiness. While the two routes derived metadata separately they disagreed about which commit a build was for — the CLI route reported the PR's merge commit, the vanilla route the real head — so one run produced builds attributed to two different SHAs. With one source they agree:

```
aspect-test:  COMMIT_SHA=8bb88722…  COMMIT_AUTHOR_EMAIL=josh@aspect.build  BUILD_USER=joshcbarnes
test:         COMMIT_SHA=8bb88722…  COMMIT_AUTHOR_EMAIL=josh@aspect.build  BUILD_USER=joshcbarnes
```

The action emits the `--build_metadata=` keys `get_build_metadata_flags` derives, plus `BUILD_USER`, `BUILD_HOST` and `COMMIT_TAG`, and runs `setup-aspect` so `aspect` is on PATH for the credential helper.

On a pull request it resolves metadata against `github.event.pull_request.head.sha` rather than `HEAD`. `actions/checkout` leaves HEAD on GitHub's synthetic merge commit, so without this every PR build in the UI is titled `Merge <sha> into <sha>` and attributed to the merge's author. The head commit is not in the shallow clone, so it is fetched at depth 1 when absent, falling back to HEAD if that fails.

It also runs `setup-aspect`, since `--credential_helper=cache.aspect.build=aspect` needs `aspect` on PATH and three of those jobs only installed Bazel.

### Which tenant

The org-level `ASPECT_API_TOKEN` authenticates as a tenant the OSS team cannot see, which is why uploads succeeded while the UI stayed empty. CI now uses `ASPECT_OSS_API_TOKEN`, with no fallback — a fallback would silently re-route builds to the invisible tenant whenever the new secret is missing.

### Coverage

| Job | Covered | Route |
|---|---|---|
| `aspect-test` | yes | `config.axl` |
| `format`, `lint`, `test` | yes | composite action |
| `format-aspect-cli`, `lint-aspect-cli` | yes | composite action |
| `integration-test` | **no**, deliberately | see below |

`integration-test` was wired and then backed out: with the remote cache enabled, `integration-test (examples/scala)` fails reproducibly. `aspect lint` reports `✅ Passed · No findings` while scalafix itself prints `src/App.scala:1: warning: Unused import`, so the lint reports reach the CLI empty. It passes on `main` in every recent run, so this is a real interaction and not a flake. Root cause is not yet established — the obvious suspect, `--remote_download_outputs=minimal` from that example's `common:ci`, is ruled out (the spawned command line carries neither `--config=ci` nor that flag). That job is byte-identical to `main` in this PR.

### Notes for review

- **Gated on `ASPECT_HAS_API_TOKEN`, not on the token itself.** `ASPECT_API_TOKEN` in the environment overrides the short-lived JWT `setup-aspect` stores, so every `aspect get` redoes the token exchange — that produced 171 credential-helper timeouts over 20 minutes before it was removed. The workflow exports only whether the secret exists. Fork PRs get no secrets, so both routes no-op and those builds are unchanged.
- **`setup-aspect`'s launcher install is deliberately ungated.** `integration-test`'s bats fixtures call `aspect lint`, so the binary must exist even without a token. Only the login is conditional.
- **Its three caches are turned off.** `setup-aspect` defaults `bazelisk-cache`, `disk-cache` and `repository-cache` on while `setup-bazel` defaults them off, so wrapping it would have silently enabled three Actions caches across 48 legs in jobs that previously cached nothing. The remote cache supersedes them — setup-aspect itself ignores them on Workflows runners for that reason.
- **`--experimental_profile_include_target_configuration` is omitted** from `aspect-bes`: Bazel 8+, and `.bazelversion` is 7.6.0; with it in, every build fails on `Unrecognized option`.
- **`BUILD_USER`, `BUILD_HOST` and `COMMIT_TAG` are added by hand** on both routes. `get_build_metadata_flags` leaves them to Bazel's workspace status, which this repo never sets — so `BUILD_USER` would default to Bazel's `whoami`, i.e. `runner`. `BUILD_HOST` uses `RUNNER_NAME`: axl exposes no hostname API, and shelling out on every CLI invocation for a marginal field is a bad trade, so both routes use the same source.
- **`@aspect//lib/build_metadata.axl`**, not `@aspect//bazel/...` as in aspect-cli — the stdlib moved, and this repo pins `2026.25.20`. A version bump will need that path updated.
- **`release.yml` is deliberately untouched.** Routing a tagged release through a remote cache means an Aspect outage can block cutting a release. Happy to wire it if reviewers want that tradeoff.
- Also drops `repository-cache: ${{ inputs.mount_bazel_caches }}` from the `test` job — this workflow has no such input, so it has always evaluated to empty.

### Test plan

Confirmed on CI, not just locally:

- All three `aspect-test` legs uploaded with full metadata — `REPO_URL`, `REPO_OWNER`, `REPO_NAME`, `USER`, `BUILD_URL`, `BRANCH_NAME`, `PR_NUMBER`, `PR_SOURCE_BRANCH_NAME`, `PR_TARGET_BRANCH_NAME` and the commit set — e.g. `app.aspect.build/i/2bcbcc6b-9bb1-4380-8640-4005bff52d94`.
- Credential helper measured at 167ms / 13ms, exit 0, for `cache.aspect.build` and `bes.aspect.build`.
- `.aspect/bazelrc` verified inert with no `--config` named.

The vanilla-`bazel` route is being exercised for the first time on this PR's latest run; the four newly-wired jobs are the confirmation.


jbedard (approved):

.github/actions/aspect-cloud/action.yaml
```diff
@@ -0,0 +1,94 @@
...
+          meta BUILD_USER "$GITHUB_ACTOR"
+          meta BUILD_HOST "${RUNNER_NAME:-}"
+          case "$GITHUB_REF" in refs/tags/*) meta COMMIT_TAG "${GITHUB_REF#refs/tags/}" ;; esac
+          if [ -n "$PR_NUMBER" ]; then
+            meta PR_NUMBER "$PR_NUMBER"
+            meta PR_ID "$PR_NUMBER"
+            meta PR_SOURCE_BRANCH_NAME "$GITHUB_HEAD_REF"
+            meta PR_TARGET_BRANCH_NAME "$GITHUB_BASE_REF"
+          fi
+        } >> "$HOME/.bazelrc"
```
jeffpignataro: Is there not a better way to handle this than a big `echo` statement in the `yaml`? Feels really awkward.

joshcbarnes: Yeah, I agree. I asked Jason for his thoughts, I think he asked you, and now it seems like the circle has been completed!

I'm open to opinions on how best to deal with this. These vars are needed in order for the UI to render well, and it's not the kind of data that belongs in an rc file. I'm currently thinking we make something like `aspect ci build-metadata` that prints the build-metadata flags?

jeffpignataro: > the circle has been completed!

Exactly.
https://media.tenor.com/Iyu3kiZCZEQAAAAe/the-circle.png

Greg just landed the `aspect setup` task group. I think `build-metadata` would make a lot of sense there. Then people could have the option of a simple workflow task. Context - https://github.com/aspect-build/aspect-cli/pull/1460

PR #1003, opened 2026-09-17
ci: connect the aspect-cli jobs to Aspect Cloud

joshcbarnes:
Points the three jobs that run Bazel through the Aspect CLI — `format-aspect-cli`, `lint-aspect-cli`, `aspect-test` — at Aspect Cloud's remote cache and BES. They run on stock GitHub-hosted runners, which have no cache of their own.

Draft: opened to exercise the one thing that cannot be validated locally — the `ASPECT_OSS_API_TOKEN` → JWT exchange.

### What it takes

`setup-aspect` already pipes its `aspect-api-token` input into `aspect auth login --with-api-token`. That login records a credential but **no endpoints**, and Aspect Cloud's built-in entry ships with none, so a token alone leaves `--remote` with nothing to wire. The endpoints come from a committed `.aspect/config.json`, which the CLI finds through `$ASPECT_WORKSPACE` — pinned to the checkout root, since the example jobs build from subdirectories. That variable selects only the deployment list, never which `.aspect/config.axl` a task loads, so each example keeps its own lint aspects.

`REMOTE_FLAG` is `--remote` when the secret is present and empty otherwise, which is every fork PR: forks get no org secrets, the login never runs, and `--remote` would reach the cache with no credential.

### Why the version bump

The pin was 2026.25.20, which predates the deployment feature — no `--remote`, no `aspect auth status`/`configure`. All 27 pins move to 2026.38.14 together so the root and the examples stay on one version. `--task-key` is deprecated there, so the three call sites move to `--task:name`.

`integration-test` is deliberately left alone: its bats tests assert on `aspect lint` output, and a BES stream adds a `Streaming build results to:` line.

### Verified locally

Run from `examples/starlark` with `ASPECT_WORKSPACE` at the repo root — the CI shape:

```
INFO: Using deployment 'aspect-cloud' for remote cache (grpcs://cache.aspect.build)
INFO: Streaming build events to grpcs://bes.aspect.build.
INFO: Streaming build results to https://app.aspect.build/i/…
INFO: 1023 processes: 3 remote cache hit, 1016 internal, 4 darwin-sandbox.
```

The example's own `config.axl` still registered its Buildifier aspect, so `ASPECT_WORKSPACE` did not hijack config loading. With `REMOTE_FLAG` empty the build runs clean with nothing Aspect-specific wired. Every `config.axl` in the repo loads under 2026.38.14.

### What this PR is meant to prove

Local runs used a browser login, not the token, so CI is the first exercise of:

- [ ] `ASPECT_OSS_API_TOKEN` exchanges for a JWT. **Note:** setup-aspect `core.warning`s rather than fails on a bad login, so check the `Exchange ASPECT_API_TOKEN for a session JWT` group explicitly — a green setup step does not mean it worked.
- [ ] That account has cache/BES write access for the org.
- [ ] The three jobs print the `Using deployment 'aspect-cloud'` and `Streaming build results to` lines.
- [ ] The version bump holds across all 27 workspaces on Linux (only `examples/starlark` was exercised, on macOS).


joshcbarnes:
## CI results — 90/90 green

The token path works. From `aspect-test (., bazel-7)`:

```
##[group]Exchange ASPECT_API_TOKEN for a session JWT
Persisted Aspect session JWT for downstream `aspect` invocations
INFO: Using deployment 'aspect-cloud' for remote cache (grpcs://cache.aspect.build)
INFO: Streaming build events to grpcs://bes.aspect.build.
INFO: Streaming build results to https://app.aspect.build/i/0017f0ce-…
INFO: 1009 processes: 557 disk cache hit, 40 remote cache hit, 412 internal.
```

`lint-aspect-cli (examples/java)` reports **985 remote cache hits**, so the cache reaches the examples too.

## Known gap: only 3 invocations reach the Web UI

`format`, `lint` and the bats `aspect lint` calls wire the remote cache but stream **no** build events, so only the three `aspect-test` jobs appear in the UI. This is not a misconfiguration here — it is aspect-build/aspect-cli#1455, which is still open and therefore absent from the 2026.38.14 pin:

> `--remote=bes` reached `build` and `test`, and silently streamed nowhere for `format`, `gazelle`, `lint`, `run`, `warming`, and `delivery`.

The cache half lands in `BazelTrait.base_flags`, which every task reads generically; the BES half sat on `BazelTrait.bes_backends`, which only `build`/`test`'s spawn path passed through. Reproduced locally: `aspect build --remote` streams BES, `aspect lint --remote` and `aspect format --remote` do not.

**Follow-up:** once #1455 merges and ships, bump the pin — no change needed in this repo. That should take this CI from 3 hosted invocations to roughly 60.

## Verification status

- [x] `ASPECT_OSS_API_TOKEN` exchanges for a JWT
- [x] The account has cache access (remote cache hits in both root and examples)
- [x] BES write access — proven by `aspect-test`; the other jobs are blocked on #1455, not on permissions
- [x] Version bump holds across all 27 workspaces on Linux
- [ ] `integration-test`'s cache use is unverified by log — bats `run` captures `aspect lint`'s output, so its INFO lines never reach the job log. Token and `--remote` are provably in the job env and the tests pass.

.aspect/config.json
```diff
@@ -0,0 +1,19 @@
+{
+  "_comment": "Generated by `aspect auth configure api.aspect.build --login=false`. Committed so CI runners know cache.aspect.build/bes.aspect.build are Aspect Cloud's; found via $ASPECT_WORKSPACE. Regenerate with that command.",
```
jbedard, line 2: Generated yet we have to check it in?

examples/cpp/.bazelrc
```diff
@@ -1,3 +1,8 @@
+# Aspect's rc for vanilla `bazel`. Generated at the repository root by
+# `aspect setup bazelrc`; regenerate there, not here. %workspace% is this
+# example, so `../..` reaches the root the examples share.
+try-import %workspace%/../../.aspect/bazelrc
```
jbedard, line 4: `import`?

examples/csharp/.bazelrc
```diff
@@ -0,0 +1,4 @@
+# Aspect's rc for vanilla `bazel`. Generated at the repository root by
+# `aspect setup bazelrc`; regenerate there, not here. %workspace% is this
+# example, so `../..` reaches the root the examples share.
+try-import %workspace%/../../.aspect/bazelrc
```
jbedard, line 4: `import`?

.bazelrc
```diff
@@ -1,3 +1,6 @@
+# Aspect's rc for vanilla `bazel`, maintained by `aspect setup bazelrc`.
+try-import %workspace%/.aspect/bazelrc
```
jbedard, line 2: `import`?

examples/shell/test/lint_test.bats
```diff
@@ -11,7 +11,7 @@ EOF
 }
 
 @test "should produce reports" {
-	run aspect lint --strategy=soft --tips:silence=add-aspect-api-token-github-actions -- //src/...
+	run aspect lint $REMOTE_FLAG --strategy=soft --tips:silence=add-aspect-api-token-github-actions -- //src/...
```
jbedard, line 14: If unset and a user runs this locally this will just be empty and work fine?

.github/workflows/workspace_status.sh
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env bash
```
jbedard, line 1: This file is ugly

jbedard (approved):

